PR #2624, opened 2023-06-26
Clean up set record and update record

humandad:
# What does this PR do?

Note: I don't think this PR is necessary for my demo, so we could hold off on merging this. I just wanted to get the code submitted.

This PR consolidates all of the uses of setRecord and updateRecord into the wire class. That way there is only one place that initiates these actions.


humandad:
closing. this is stale.

PR #3330, opened 2023-10-23
Nicer validation errors when deploying yaml

Wesselvdp:
# What does this PR do?

Old:
image

New:
image

Maybe we can get rid of the `Field 'Definition' failed YAML schema validations:` entirely?
I'm not sure where the trailing `:` is coming from after the first error block:
image

# Testing


humandad:
this is stale. closing.

PR #3387, opened 2023-11-02
Add nav tile and icon tile components

humandad:
# What does this PR do?

Attempts to parse componentvariant properties serverside. I just wanted to get this code out there. It may not be the right approach.


humandad:
this functionality was added in a more recent PR.

PR #3405, opened 2023-11-06
feat: impersonation for workspace

abelcloudmasters:
# What does this PR do?
- new component 
- Impersonation is now close to the view.
- there were bugs on the route reload so I could not use it in the view builder.
- removes the old impersonation in the profile section.

# Testing


humandad:
This is pretty stale. Closing for now.

PR #3493, opened 2023-11-21
[PARKED] fix: FIELD of type USER should show the right users in workspace

abelcloudmasters:
# What does this PR do?

Some context:

- in the workspace context we give full access to all collections using the GetAdminPermissionSet function, since the variable ViewAllRecords is true we skip the AccessCheck.

- I force the component to always use the "site" context so we get the records that we expect by using the function deleteWorkspace.

Basically, we never query for users from a WS context, since doing this will give us all users.
I think we need to add more Record Challenge Tokens to get also the team members etc. (doing this in my local testing it out)

# Testing


humandad:
closing since this has gotten stale.

PR #3761, opened 2024-01-29
[DRAFT] feat: install bundles bot

abelcloudmasters:
# What does this PR do?

Now you can install bundles from the prod bundle store.

# Testing

go to the bundle store and try to install prod bundles.


humandad:
Went a different approach here.

PR #3781, opened 2024-01-30
update nx to 17

humandad:
# What does this PR do?

Updates nx to version 17.

All changes here were automated and the result of running the following commands.

1. `nx migrate latest`
2. `npm install`
3. `nx migrate --run-migrations`
4. Deleted the `migrations.json` file

# Testing

Everything should work the same as before.


zachelrath (approved):
Looks good to me, so far things seem to be working

PR #3790, opened 2024-01-31
Omit empty builtin user fields

humandad:
# What does this PR do?

I came across this while inspecting some information that was stored in redis. There were a lot of null keys for these fields since they weren't queried. Since these are part of every object, it's nice to omit them if they are empty.

# Testing

Everything should work the same as before.


zachelrath (approved):
Nice. We could probably do some omitempty fields on the built-in User struct as well.

What happens if you don't do the change in tags.go ?

PR #3842, opened 2024-02-08
fix(bots): Bundled bot code not invalidated immediately after site bundle update

### LINKED ISSUE #3841: Bundled Bot code not invalidated after bundle deploy
# Steps to replicate

1.  Create a Bot
2. Add it to a Bundle, and package a version.
3. Update a site to use that Bundle.
4. Modify the Bot's code.
5. Package a new Bundle version.
6. Update a site to use that Bundle.
7. EXPECTED --> the new Bot code would be used immediately.
8. ACTUAL ---> it takes 15 minutes or so for the new Bot code to be used.

zachelrath:
# What does this PR do?

closes #3841 

Previously, in Site mode, we were only caching Bot programs using the Bot's key, which meant that, for 5-15 minutes after you updated a Site to a new bundle, the previous bot code would still be used.

Now, we use the context site's app name and bundle version in the Site-mode Bot program cache key, to ensure that net-new Bot code will be immediately used after the site is updated to a new Bundle version.

# Testing

See testing steps in the issue description.

### DIFF
--- a/apps/platform/pkg/bot/jsdialect/jsdialect.go
+++ b/apps/platform/pkg/bot/jsdialect/jsdialect.go
@@ -53,18 +53,27 @@ func (b *JSDialect) hydrateBot(bot *meta.Bot, session *sess.Session, connection 
 	if _, err := bundle.GetItemAttachment(buf, bot, b.GetFilePath(), session, connection); err != nil {
 		return err
 	}
-	bot.FileContents = string(buf.Bytes())
+	bot.FileContents = buf.String()
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


apps/platform/pkg/bot/jsdialect/jsdialect.go
```diff
@@ -98,8 +107,8 @@ func RunBot(bot *meta.Bot, api interface{}, session *sess.Session, connection wi
 		return err
 	}
 	time.AfterFunc(time.Duration(getTimeout(bot.Timeout))*time.Second, func() {
+		// Interrupt native Go functions
```
zachelrath, line 110: another Go lint fix, the return was redundant and unnecessary

humandad (approved):

PR #3863, opened 2024-02-17
refactor: Start to extract record level access enforcement out of Postgresio adapter

zachelrath:
# What does this PR do?

1. Drops total queries performed during Uesio Integration tests by 25%, by reducing the number of unnecessary record level access queries that Uesio makes by doing a better job of checking to see whether there are collections in a Load request actually needing record level access checks. 
2. Begins to refactor Postgres Load adapter to be able to function against an external PG database other than Uesio's multi-tenant, single-table structure.
3. Adds a new IdField field on Collections, which will be used to identify metadata related to the "id" field of external Postgres Collections.

## Queries: Master

```
{"queryStats":{"totalWorkspaceQueries":2113,"totalQueries":15841}}
```

## Queries: This branch
```
{"queryStats":{"totalWorkspaceQueries":2103,"totalQueries":12119}}
```

## Delta: 25%

# Testing

There should be no functional changes to the app at this point, only reduced queries.

### DIFF
--- a/apps/platform/pkg/adapt/postgresio/load.go
+++ b/apps/platform/pkg/adapt/postgresio/load.go
@@ -170,7 +170,7 @@ func getAliasedName(name, alias string) string {
 func (c *Connection) Load(op *wire.LoadOp, session *sess.Session) error {
 
 	metadata := c.metadata
-	userTokens := session.GetFlatTokens()
+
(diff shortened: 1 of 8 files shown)


humandad (approved):
Looks good. Thanks! 🙏

PR #3864, opened 2024-02-18
fix: Support SELECT fields in Chart labels/categories, do count on non-numeric value fields

zachelrath:
# What does this PR do?

1. Allow SELECT fields to be used in Charts by treating them like TEXT fields.
2. Performs a count on non-numeric value fields.
3. Adds unit tests for the `aggregate` function.

# Testing
1. Add a chart with a series whose Category Field is a SELECT, and whose Value Field is uesio/core.id. Verify that the values for each Select value are counted.

# Screenshots

Screen Shot 2024-02-17 at 10 07 45 PM

### DIFF
--- a/libs/ui/src/wireexports.ts
+++ b/libs/ui/src/wireexports.ts
@@ -65,7 +65,6 @@ export type {
 	ValueConditionState,
 	ViewOnlyField,
 	ViewOnlyWireDefinition,
-	Wire,
 	WireConditionState,
 	WireDefinition,
 	WireDefinitionMap,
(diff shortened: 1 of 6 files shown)


libs/ui/src/wireexports.ts
```diff
@@ -75,4 +74,4 @@ export type {
 	WireRecord,
 }
 
-export { isValueCondition, isGroupCondition, isParamCondition }
```
zachelrath, line 78: this allows for use of the Wire class constructor, in addition to the Wire type.

humandad (approved):

PR #4062, opened 2024-07-08
send email to admin when a new user signup using google

abeljimo:
# What does this PR do?

This emails the studio admin when a new user signs up using the Google method.

# Testing

You can test this locally, by adding the Google and SendGrid keys and ensuring that the config value signup_notify_email is set.

Then just create a new user using the Google method.


humandad (approved):
Looks good to me. Thanks @abeljimo

PR #4364, opened 2024-12-20
run cli from dist directory

### LINKED ISSUE #4353: bug: build scripts (e.g., `npm run build-all`) should use `dist` versions of uesio
The docs mention creating a [symlink](
image
) for `uesio` cli and some of the build scripts (e.g., [libs/apps/uesio/aikit](https://github.com/ues-io/uesio/blob/master/libs/apps/uesio/aikit/project.json#L13)), although not all (e.g., [apps/platform](https://github.com/ues-io/uesio/blob/master/apps/platform/project.json#L25)), rely on `/usr/local/bin` while others use the binaries in `dist` directory.

When building the product, all usage of binaries that are built as part of the build process should be used in subsequent build steps to complete product build.

techfg:
# What does this PR do?

Ensures that build scripts use `dist/cli/uesio` consistently.

I believe I located all places that non-relative references existed but a thorough review of the build process should be performed to ensure I didn't miss anything 😄 

Resolves #4353

# Testing

`npm run build-all` works as expected without requiring symlink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,8 @@ Uesio is a **low-code** application development platform.
     -   `jq` (for JSON manipulation in Shell): `brew install jq`
     -   `wget` (for fetching URLs): `brew install wget`
 -   Start dependencies [here](#dependencies).
--   Create a symlink for the Uesio CLI into your bin (NOT an alias, which won't work with `nx`):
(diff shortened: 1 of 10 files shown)


humandad (approved):
This looks good to me. I made a small addition to this PR that removes the symlink setup in the github builds as it's no longer necessary. Looks like builds are passing fine without it.

PR #4370, opened 2024-12-21
fix: view component not loading correctly

### LINKED ISSUE #4369: bug: view component displays `Could not get View Def` error in builder
There are a few problems with the display of a `View` component within a `View`:

1. Create an app
2. Create a view named `child`, add a `text` component, save it
3. Create a view named `parent`, add a `view` component to it

Actual Behavior
The `view` component displays an error `Could not get View Def`

Expected Behavior
The `view` component is added to the view and no error is displayed  (ideally "empty components" have some type of visual indication of where they exist but that is a separate issue that needs to be tracked).

4. Change the `View` properties to reference `child` view

Actual Behavior
The `view` component displays an error `Could not get View Def`

Expected Behavior
The `view` component displays the `child` component

5. Save the page & refresh

Actual = Expected
The `child` component displays

6. Removing the reference to `child` view from the `view` properties 

Actual Behavior
The `view` component displays an error `Could not get View Def`

Expected Behavior
The `view` component is updated to "empty space" (ideally "empty components" have some type of visual indication of where they exist but that is a separate issue that needs to be tracked).

techfg:
# What does this PR do?

Ensures that a view component does not display an error when no view is selected and properly displays when selected.

Resolves #4369

# Testing

Manual testing of scenarios described in #4369 including monitoring for number of network calls resulting from memoized result.

@humandad - Couple of things specifically for review beyond general:

1. If the network call fails, error displays in console due to unhandled rejection occuring in useEffect which isn't ideal.  LIkely better to have the component itself display an error possibly.  Not sure what the standard is in the codebase for these types of situations.  Thoughts?
2. Regarding the Wires component, I tried to find & identify how FunctionComponents are defined in other areas and follow similar/same pattern.  It's a little overkill since Wires will only ever be used by View but have a look and let me know if it's what your expecting.
3. Plan on adding tests after we talk about approaches but, as discussed, getting that done shouldn't hold-up review & merge of this.

Thanks again for all your help in getting this fixed!!

### DIFF
--- a/libs/apps/uesio/builder/bundle/componentpacks/main/src/components/paramsfield/paramsfield.tsx
+++ b/libs/apps/uesio/builder/bundle/componentpacks/main/src/components/paramsfield/paramsfield.tsx
@@ -51,14 +51,12 @@ const ParamsField: definition.UC<ParamsFieldDefinition> = (props) => {
 		}
 	}
 
-	if (!view) return null
-
 	const paramsDef = api.view.useViewDef(view)?.params as Record<
 		string,
 		Record<"type", string>
 	>
 
-	if (!paramsDef) return null
+	if (!view || !paramsDef) return null
 
 	const params = record.getFieldValue(fieldId)
 
--- a/libs/ui/src/bands/view/operations/load.ts
+++ b/libs/ui/src/bands/view/operations/load.ts
@@ -23,7 +23,7 @@ const runEvents = async (
 	}
 }
 
-const useLoadWires = (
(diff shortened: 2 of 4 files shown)


humandad (approved):
Barry, this looks great! The only change I made was to rename the component to `WiresAndEvents` because I realized the hook also handles the events node on a view def.

I'm ok with how the errors are thrown if the network goes down or something. Also the formatting of the react component looks good to me.

Thanks!

PR #4404, opened 2025-01-04
fix: do not ignore dotfiles when zipping local files

### LINKED ISSUE #4400: bug: `deploy` ignores files that start with `.` (dotfiles)
During a deploy, the contents of the local file system is zipped and deployed to the server.  However, if the filename of a file on disk starts with a `.` (period/dot), the file is not uploaded to the server and [intentionally ignored](https://github.com/ues-io/uesio/blob/master/apps/cli/pkg/zip/zip.go#L31).

The only way to modify certain meta data items that start with a `.` is to use the Web UI which does allow uploading files that start with `.`.  Additionally, the `retrieve` command will pull files that start with a `.` down. 

If there is a valid reason to ignore files that start with a `.`, then the web UI should be updated to not allow this and documentation reflecting the limitation.  Alternatively, the CLI should not ignore files that start with a `.`.

Note - This rule applies to all meta data types & meta data items.  Some items, such as Views, do not allow `.` characters anywhere in the name, however some (e.g., Files) do.  All metadata items should be consistent in terms of their ability to retrieve/deploy via Web UI and/or CLI.

techfg:
# What does this PR do?

Changes behavior from ignoring all dotfiles (files that start with `.`) to not ignoring.  There are certain meta data types that allow files that start with `.` (e.g., Files).  If a file gets zipped and is invalid for a given metadata type, server side validation when processing the zip should detect and fail the operation.

# Testing

No current tests for any portions of the CLI in this area so manually tested to ensure that files that start with `.` are zipped and sent to server and changes reflected.

Resolves #4400

### DIFF
--- a/apps/cli/pkg/zip/zip.go
+++ b/apps/cli/pkg/zip/zip.go
@@ -28,9 +28,6 @@ func ZipDir(localPath string) io.Reader {
 				return nil
 			}
 			file, err := os.Open(path)
-			if strings.HasPrefix(file.Name(), ".") || strings.Contains(file.Name(), "/.") {
-				return nil
-			}
 			if err != nil {
 				return err
 			}
--- a/apps/cli/pkg/zip/zip.go
+++ b/apps/cli/pkg/zip/zip.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/thecloudmasters/uesio/pkg/fileadapt/localfiles"
 )
 
 func ZipDir(localPath string) io.Reader {
@@ -25,6 +27,9 @@ func ZipDir(localPath string) io.Reader {
 				return err
 			}
 			if info.IsDir() {
+				return nil
+			}
+			if localfiles.ShouldIgnoreFile(info.Name()) {
 				return nil
 			}
 			file, err := os.Open(path)
--- a/apps/platform/pkg/fileadapt/localfiles/localfiles.go
+++ b/apps/platform/pkg/fileadapt/localfiles/localfiles.go
@@ -7,11 +7,22 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/thecloudmasters/uesio/pkg/types/file"
 	"github.com/thecloudmasters/uesio/pkg/types/wire"
 )
+
+// Skip .DS_Store files
+var LocalFileIgnore = []string{".DS_Store"}
+
+// For local files, returns true if the file should be ignored during processing, false otherwise
+func ShouldIgnoreFile(fileName string) bool {
+	return slices.ContainsFunc(LocalFileIgnore, func(name string) bool {
+		return strings.EqualFold(name, fileName)
+	})
+}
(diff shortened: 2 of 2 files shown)


humandad (commented):
I think the main issue here that macs create a `.DS_Store` file in every directory if it's opened in "finder". I'm not sure if windows has something like this too. I want to avoid having these files end up as part of componentpacks, files and other metadata types that can take any files. We could have a list of just those files specifically and comment it better about what we're trying to accomplish though.

techfg:
> I think the main issue here that macs create a `.DS_Store` file in every directory if it's opened in "finder". I'm not sure if windows has something like this too. I want to avoid having these files end up as part of componentpacks, files and other metadata types that can take any files. We could have a list of just those files specifically and comment it better about what we're trying to accomplish though.

This makes sense but the question is, how far do we take this vs. just sucking all the files in which will be ignored anyway.  Windows doesn't does put a .thumbs if the directory has images and if viewed through Windows Explorer I believe.  Maybe the compromise is implementing a .packignore type of concept (either exposed or just internal in the code) where it forcibly ignores known files and lets everything else through?

humandad:
I would be fine with just a hard-coded list of intentionally ignored files and us filtering them out. As long as `.DS_Store` is on that list, I'm happy.

techfg:
Cool - anything else beyond .DS_Store or should we just start with that and expand if/when needed?

humandad:
I'm fine with sticking with .DS_Store for now. While we're at it we could switch to the newer, more performant `WalkDir` function as well. https://pkg.go.dev/path/filepath#WalkDir

techfg:
Yeah, considered the WalkDir change amongst a few other changes.  I think we would actually want to change to [fs.WalkDir](https://pkg.go.dev/io/fs#WalkDir) ideally (e.g. testability) but `filepath.WalkDir` would be next best option.  Was trying to avoid changing too much in one go so think it best to revisit some broader changes for separate PR 😄

techfg:
Btw, looks like [.DS_Store](https://github.com/ues-io/uesio/blob/de602f8ef77cd9fe43a05e89dae7009feedb3d46/apps/platform/pkg/fileadapt/localfiles/localfiles.go#L58) is handled as a one-off case in this situation along with using WalkDir :)

I'll combine the logic so that both Walks use an ignore list.

humandad:
> Btw, looks like [.DS_Store](https://github.com/ues-io/uesio/blob/de602f8ef77cd9fe43a05e89dae7009feedb3d46/apps/platform/pkg/fileadapt/localfiles/localfiles.go#L58) is handled as a one-off case in this situation along with using WalkDir :)
> 
> I'll combine the logic so that both "Walks" use an ignore list.

Sounds good. Also I agree about the separate PR for WalkDir.

techfg:
@humandad  -

Added in a utility function and slices of file names that should be ignored during local file processing.  Not sure if the slice & utility function are in the best place file wise but couldn't find a "path" type of utility file.  Let me know if you'd like me to relocate the slice & func and where to put it.

Thanks!

humandad (approved):
This is great. Thanks!

Issue #4393, opened 2025-01-03
bug: cli commands (e.g., `retrieve`) fail when not executed from workspace root directory


techfg:
When executing the cli, commands such as `retrieve` fail because filepaths are relative to `cwd` instead of incorporating the `-d` parameter where needed.

For example, when a workspace exists in `/apps/myapp`, executing `uesio retrieve` from `apps` directory fails:

```bash
cd /uesioapps
./uesio app clone -d ./myapp
./uesio workspace retrieve -d ./myapp
```
```text
Error: unable to read file at path: bundle/bundle.yaml
```

All commands should support the CLI being executed from anywhere, incorporating/using the `-d` flag as needed.  As a general rule, resolving paths to absolute paths and then using relative pathing to join to base should eliminate most/all of these types of issues.

PR #4405, opened 2025-01-04
fix: log message when `-d` option is not specified

### LINKED ISSUE #4402: bug: commands emit incomplete message when `-d` is not specified
When the `-d` flag on various commands is not specified, the message that the commands emit when starting contains an "empty" directory reference.

For example:

```bash
./uesio retrieve
Retrieving metadata from studio into  ...
```

```bash
./uesio retrieve -d .
Retrieving metadata from studio into current directory ... 
```

In the first example, the text is "into  ..." with no "directory" included in the message.

Regardless of whether or not `-d` is specified, the message should indicate the correct directory for commands that emit messages and have a `-d` flag.

techfg:
# What does this PR do?

When `-d` is not specified for `retrieve` or `deploy` commands, ensure log message contains "current directory"

> [!NOTE]
> There are several issues with the way that paths are resolved when `-d` is specified.  For example, if `./` is specified, retrieve fails with `Error: illegal file path: bundle/views/community_detail.yaml` error.  This PR only addresses the specific case around not specifying `-d`.  Separately, the entire code base should be reviewed/adjusted to ensure proper handling of `-d` depending on the approach taken in #4393.

# Testing

No existing tests so manual testing performed to ensure proper message emitted.

Resolves #4402

### DIFF
--- a/apps/cli/pkg/command/workspace/deploy.go
+++ b/apps/cli/pkg/command/workspace/deploy.go
@@ -16,7 +16,7 @@ import (
 func Deploy(sourceDir string) error {
 
 	sourceDirDescription := sourceDir
-	if sourceDir == "." {
+	if sourceDir == "." || sourceDir == "" {
 		sourceDirDescription = "current directory"
 	}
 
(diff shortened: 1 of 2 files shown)


humandad (approved):
Looks good to me.

PR #4410, opened 2025-01-04
fix: display slot name when label not provided

### LINKED ISSUE #4409: bug: Slots do not display their name in component index
When adding components that contain slots to the canvas, there is no way to know which Slot is which because the name of the slot does not display anywhere.

For example, adding a Card displays the following in the component index:
image

techfg:
# What does this PR do?

If a slot does not specify a `label` property, display the `name` of the slot so that the component index contains something meaningful rather than `Slot` so that the user knows which slot is which.

# Testing

No current test coverage so manual testing of the builder on a component that does not contain labels for slots (e.g., Card).

Resolves #4409


humandad (approved):
yeah, this is definitely better. Thanks!

PR #4411, opened 2025-01-06
chore: improve performance of zipping files

techfg:
# What does this PR do?

Switches from `filepath.Walk` to `filepath.WalkDir` to improve performance of zipping files.

This is an intermediate solution as the final solution should migrate to using [fs.WalkDir](https://pkg.go.dev/io/fs#example-WalkDir) to improve testability.  Making the change to `fs.WalkDir` requires a more invasive set of changes so for now, just picking up perf improvements of `filepath.WalkDir` vs. `filepath.Walk`.  Migrating to `fs.WalkDir` should be completed as part of a larger effort required to improve overall testability of the codebase.

# Testing

No current tests so manual deploy performed via CLI.

### DIFF
--- a/apps/cli/pkg/zip/zip.go
+++ b/apps/cli/pkg/zip/zip.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


humandad (approved):
Thanks!

PR #4425, opened 2025-01-07
Adds support for custom fonts

humandad:
# What does this PR do?

1. Adds a new metadata type called "font". This can be used to add fonts to bundles.
2. Removes the hard-coded fonts and all of the build scripting around them.
3. Adds `roboto` and `material_icons` fonts to the `uesio/io` package.
4. Removes the built-in `Gosha`, `Lato`, and `Viga` fonts. These should be added in other packages that require them. The Lato and Viga fonts were added for an old project that is no longer active. The Gosha font should be added locally to any app that needs to use it.

Removes font face observer hack - This was to fix a bug in safari, but it seems like I can no longer reproduce it.

# Notes

In its current form, this just adds all CSS for bundled fonts to the dependencies for every route load. This means that the css for all accessible fonts will be added to route load. This does not mean that all fonts will be loaded for each route load though. That is up to the browser to request the font if it encounters it. We could definitely add more metadata around fonts to control preloading/etc. We could also allow routes, themes or views to explicitly specify the fonts they want to load, but I think it's ok for now to just load them all as dependencies.

Also, this PR does not include a studio UI for working with fonts. That would be added in a subsequent PR. I do think this gets us on the right path for supporting custom fonts though, and is mostly backwards compatible, (Aside from removing the Gosha font)

# Testing

1. Verify that fonts load in the studio appropriately.
2. Test in safari and an iphone to determine if fonts are loading correctly.
3. Add a new font to an app and verify that you can deploy/retrieve and use that font.


techfg (commented):
Per our discussion, looks good!

PR #4446, opened 2025-01-14
[WIP] Fix/issue 4437 builder component marker

### LINKED ISSUE #4437: bug: unable to identify/select component that has no content in builder canvas
Currently, when a component has no content, it is "invisible" on the builder canvas.  It is available in the Component Index but even when clicking on it on the index, you cannot identify where it is on the canvas.

There needs to be a way to visualize the component on the canvas even when there is no content in it.

techfg:
# What does this PR do?

Places an outline on all top-level components in the builder canvas so that they can be identified visually even they do not render anything within.

Currently a WIP to solicit input/feedback before finalizing.

A few thoughts:
1. This is just an initial concept to solicit feedback on the best way to achieve the desired outcome per #4437 
2. I chose a `min-height: 1.25rem` because when a component is selected, it is given an outline of `8px` width which equates to 1rem so choosing 1.25rem provides a small amount of whitespace between the red "selected" outline
3. If a component itself returns null, the "empty" classes ensure the contents element itself has a visual border (it uses display:contents so it doesn't have block dimensions itself)
4. I used the same border style (dashed) and color (slate-300) as we use elsewhere so possibly this should be a different style and/or color
5. I wrote the classes such that they can be easily rendered out planning ahead for the new "Action Bar" in the builder that could contain a "Hide Markers" button.  The marker itself could be rendered out conditionally in one of two ways:
    1.  The button in the action bar would trigger different conditional behavior within buildwrapper.tsx
    2. The button could put a class on the top level element of the builder and then the marker classes could be prefixed with `:not(.hide-marker) ...`
6. One additional thing to consider is whether or not each slot within a component should be given the same/similar type of marker so that its easy to visually identify each slot vs just the entire component.  Additionally, currently, there is no way to drag/drop a component in a slot on the canvas if it doesn't render anything (e.g., card footer), the only way to put something in a slot is via component index or yaml - thoughts?
     - **EDIT** Actually, it doesn't look like drag-n-drop works within a slot even if contains components that are visible on the canvas.  For example, if a Card has 5 text components in the Content slot, I can't drag another one to the Content slot in the canvas, seems only way is to add via component index or yaml.  Do slots support drag-n-drop currently? 

# Testing

Tested in the builder UI with components that render something visual, doesn't render anything visual but has an element itself, renders null/nothing.

Resolves #4437


techfg:
Closing this in favor of https://github.com/ues-io/uesio/pull/4609 which made some improvements for #4437.  More improvements still required which will be addressed individually.

Issue #4432, opened 2025-01-10
feat: improve accessibility to meet WCAG standards


techfg:
Currently, ues.io (website, studio, runtime, etc.) do not meet WCAG standards in many situations.  For example:

1. Contrast ratios
2. Image alternate text
3. Form element labels
4. Keyboard access for scrollable regions
5. Links must have discernible text (e.g., icon buttons)
6. Nested interactive controls
7. Touch targets at least 24px
8. Landmarks 

A decision needs to be made on which WCAG standard should be met (e.g., 2.1, 2.2, etc.) and then updates made to achieve.

Reference:
1. https://www.w3.org/WAI/standards-guidelines/wcag/
2. https://www.etsi.org/human-factors-accessibility/en-301-549-v3-the-harmonized-european-standard-for-ict-accessibility
[latest version](https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf)

PR #4451, opened 2025-01-15
fix: dialog dismiss handling

techfg:
# What does this PR do?

Updating to `@floating-ui/react@0.27.3` in PR #4449 exposed an issue with the way that dismissing a dialog was being handled.  In v0.27.3, new functionality was added to [referencePress](https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact%400.27.0) to handle native clicks.  This meant that click events would be processed and close the dialog when not intended.  For example, when clicking anywhere within the dialog "whitespace", on a "select" field, etc.

This behavior was reproduceable with v0.26.8 if the [e.stopPropogation()](https://github.com/ues-io/uesio/compare/fix/dialog-dismiss-handling?expand=1#diff-f433cef287b5d099ddc5915d2009b888830aa517f6984d6a630255fabbed99d2L81) was removed since essentially stopping propagation was solving the underlying issues with the way the floating elements were configured.

In short, this PR resolves the following issues:

1. Eliminates the "reference element" within the dialog itself - Previously, the reference element was being assigned to the FloatingOverlay.  However, reference elements are intended to be assigned to the element that "triggers" the floating element to appear (e.g., a button that you click to launch the dialog).  Reference elements are how the library ties aria-labels, roles, etc. together between the element that triggers the floating element and vice-versa.  In short, by assigning the overlay as the reference element itself it was behaving like the triggering element and the overlay which is why stop propogation was required.  As a separate issue (see #4432), a proper solution needs to be implemented to associated triggering elements to dialogs for accessibility purposes - this could possibly be done via a prop to Dialog.
2. Assigns the prop `closeOnOutsideClick` to `outsidePress` to ensure clicking outside of the dialog itself (e.g., on the overlay) will dismiss the dialog based on the value of the prop.
3. Adds a FloatingPortal - While not directly associated with the problem, wrapping the floating elements in a portal ensures that the elements are placed outside of app root to ensure any container clipping (e.g. overflow-hidden), etc. does not impact the dialog itself.  This may help with some of the display issues of the dialog in the builder but either way, is the recommended approach per the docs.

# Testing

`tests-all` passes locally and ci & e2e should cover as well.

### DIFF
--- a/libs/apps/uesio/io/bundle/componentpacks/main/src/utilities/dialogplain/dialogplain.tsx
+++ b/libs/apps/uesio/io/bundle/componentpacks/main/src/utilities/dialogplain/dialogplain.tsx
@@ -1,5 +1,6 @@
 import { definition, styles } from "@uesio/ui"
 import {
+	FloatingPortal,
 	FloatingFocusManager,
 	FloatingOverlay,
 	useDismiss,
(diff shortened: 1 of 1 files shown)


humandad (approved):
Thanks, Barry. This is so much better. I was definitely using floating ui wrong before.

PR #4452, opened 2025-01-15
New builder UI

humandad:
# What does this PR do?

Starts the process of addressing feedback on builder UX.

Screenshot 2025-01-15 at 10 37 36 AM

1. Adjusts most of the builder panels to a "Dark Mode" styling. (In the future this could be based on the user's preferences)
2. Cleans up styling on the Component index to look less cluttered.
3. Adds a "Build Bar" that floats in the canvas. This is visible in both build and preview mode when working in a workspace.
4. Removes the padding around the canvas and dashed lines. (These were causing more problems than they were solving.)
5. Removes the breadcrumb area and uesio logo from the top left of the screen and moves it to the build bar.
6. Removes the profile area from the bottom left of the screen and moves it to the build bar.
7. Allows actions to be taken in the canvas if the user is holding down the shift key.

STILL UNFINISHED
There are some things that are not done yet here, but I want to merge this so these things can be worked in parallel.

1. Build Bar Handle - this does not work. It is intended to be a handle for moving the build bar around.
2. Build Bar Tools - The Manage Builder Panels, Device Sizing, and Collapse Build Bar buttons don't work.
3. Hotkey Work - Work needs to be done to unify and test all hotkeys in the builder. These should work in windows/mac/linux and be consistent and discoverable by the user.


techfg:
Love seeing this Ben, great work!!

PR #4496, opened 2025-01-23
Suppress window scrollTo warnings

humandad:
# What does this PR do?

Just an idea for suppressing the window.scrollTo warnings in jest.

Based on this stackoverflow post.
https://stackoverflow.com/a/57312218

@techfg feel free to merge if you like this idea or close.

Thanks!


techfg (approved):
Adjusted to use a jest mock in case it ever needs to be asserted as called, etc.  Also changed to `setupFilesAfterEnv` config option which doesn't seem to make any difference but ensures the test framework is fully loaded before stubbing.

PR #4517, opened 2025-01-29
Fix/workflow permissions

techfg:
# What does this PR do?

Fix non master workflow SHAs for nx affected

# Testing

ci and e2e will cover


techfg:
No longer required. Replaced by new CI workflow

PR #4555, opened 2025-02-04
Add retrieve types only command

humandad:
# What does this PR do?

Adds an option to `uesio retrieve` to only retrieve app specific type definitions.

`uesio retrieve --onlyTypes`

### DIFF
--- a/apps/cli/pkg/cmd/workspace.go
+++ b/apps/cli/pkg/cmd/workspace.go
@@ -26,7 +26,7 @@ func init() {
 		Long:  "Retrieves all metadata from a remote workspace to the local directory",
 		Run:   workspaceRetrieve,
 	}
-	workspaceRetrieveCmd.Flags().BoolP("onlyTypes", "t", false, "Only retrieve types")
+	workspaceRetrieveCmd.Flags().Bool("onlyTypes", false, "Only retrieve types")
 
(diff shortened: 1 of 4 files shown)


apps/cli/pkg/cmd/workspace.go
```diff
@@ -26,6 +26,7 @@ func init() {
 		Long:  "Retrieves all metadata from a remote workspace to the local directory",
 		Run:   workspaceRetrieve,
 	}
+	workspaceRetrieveCmd.Flags().BoolP("onlyTypes", "t", false, "Only retrieve types")
```
techfg, line 29: Should we drop the shorthand `t` for now or at least until we know this is going to stick.  Thinking `t` might be common for other types of things down the road and `onlyTypes` is more than likely a less commonly used flag.  Thoughts?

apps/cli/pkg/cmd/workspace.go
```diff
@@ -72,13 +73,17 @@ func init() {
 		Short: "uesio retrieve",
 		Run:   workspaceRetrieve,
 	}
+	oldRetrieveCommand.Flags().BoolP("onlyTypes", "t", false, "Only retrieve types")
```
techfg, line 76: Same as above regarding `t`

techfg (commented):
Looks good, only one main question

PR #4569, opened 2025-02-05
Treat if merge with false string as false

humandad:
# What does this PR do?

We have an `$If{}` merge that accepts a three-part value. First a condition, and then an if_true_merge and finally, and if_false_merge.

Since we merge into template strings in multiple stages, merges in the condition could result in the string "false". These conditions should still trigger the "false" condition, not the "true" condition.

Example Merge:

`$If{[$Prop{collapsed}][trueValue][falseValue]}`

If `$Prop{collapsed}` resolves to the string "false" we want the falseValue to be the final result of the if merge.

### DIFF
--- a/libs/ui/src/context/merge.ts
+++ b/libs/ui/src/context/merge.ts
@@ -250,7 +250,8 @@ const handlers: Record<MergeType, MergeHandler> = {
     } catch {
       throw InvalidIfMergeMsg
     }
-    return context.merge(parts[0])
+    const conditionResult = context.merge(parts[0])
+    return conditionResult && conditionResult !== "false"
       ? context.merge(parts[1])
       : context.merge(parts[2])
   },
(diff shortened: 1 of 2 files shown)


libs/ui/src/context/merge.ts
```diff
@@ -250,7 +250,8 @@ const handlers: Record<MergeType, MergeHandler> = {
     } catch {
       throw InvalidIfMergeMsg
     }
-    return context.merge(parts[0])
+    const conditionResult = context.merge(parts[0])
+    return conditionResult && conditionResult !== "false"
```
techfg, line 254: should we explicity check for any/all of the following as well?

"0"
"null"
"nil"
"no"

techfg: sounds good!

PR #4610, opened 2025-02-21
update nx minor version to 20.4.6

humandad:
# What does this PR do?

updates nx to 20.4.6

I was trying to remove some vulnerabilities from npm install, but it seems like nx still has some work to do to fix those.

https://github.com/nrwl/nx/issues/30016

Also, I had to change the overrides on typescript from `$typescript` to the version of typescript because of some kind of bug in npm. I updated my npm to the latest version and then started getting the error.

https://github.com/npm/cli/issues/5730

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,7 +74,7 @@ jobs:
       - name: Build and test
         env:
           UESIO_DEV: "true" # required to ensure packui gets built
-        run: npx nx affected -t build test --configuration=ci --parallel=5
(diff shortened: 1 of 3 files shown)


techfg:
@humandad - for the npm issue,  you can change to 5.7.3, `npm install` then change back to `$typescript` and `npm install` should work. Better to keep this at $typescript and work around npm issue in this manner when needed.

humandad:
Ok. I'll do that. Thanks!

techfg:
Closing in favor of #4612

PR #4621, opened 2025-02-24
fix: eliminate default variant magic string

### LINKED ISSUE #4433: bug: `default` variant magic string causes inconsistency between builder & runtime behavior
Currently, if a component does not have a `defaultVariant` property defined, the code will [auto("magically") assign](https://github.com/ues-io/uesio/blob/master/libs/ui/src/component/component.tsx#L268) `${keyNamespace}.default` to the value for defaultVariant.

This approach has pros/cons.  Generally speaking, magic strings should be avoided, however in this case there is something to be said for auto applying a default of `default` for variants.

Need to review, evaluate and decide what the preferred behavior is here: 1) Default to a `default` if a `defaultVariant` is not specified or; 2) Don't.

**Edit**
After further reviewing this, there is an issue that presents in "runtime" mode that does not present in "builder" mode due to the magic string.  In short, in builder mode, all component variants are loaded server side so that they are available.  Given this, when the "magic string" of `${keyNamespace}.default` is applied, because the variant was loaded, it can be used.  However, in runtime mode, only the variants that are explicitly referenced in yaml are loaded so even with the magic string, the variant won't be loaded (unless it is referenced elsewhere in another component).  This leads to "runtime" mode not having the styles applied whereas "builder" mode does.  For example, if "titlebar" component is used, the "subtitle" text will be styled based on the "default" component variant for "titlebar" when in builder mode.  However, in runtime (e.g. preview), the styles are not applied (because the default variant isn't referenced explicitly anywhere else on a view that only contains a single titlebar).

In short, the "magic" string is "helpful" in "builder" mode but causes inconsistency in behavior in runtime mode.  To resolve, there are two options:

1. Update server side code to load contain similar logic to what client side "magic string" code does and load in "default" variant
2. Eliminate magic string completely

Option 1 has shortcomings - for example, there are cases where there is no "default" variant for a component (e.g., a google analytics component).  Given there will never be a 100% guaranteed way to know if there is a ".default" variant and if it should/shouldn't be used, the decision is to eliminate magic string completely and require that if a component should apply a default variant, it should be explicitly included in the component definition's `defaultVariant` property.

techfg:
# What does this PR do?

Eliminates the "magic string" for default variants for components as mentioned in #4433.

Resolves #4433

# Testing

Manual testing combined with e2e & ci.

### DIFF
--- a/libs/apps/uesio/io/bundle/components/field.yaml
+++ b/libs/apps/uesio/io/bundle/components/field.yaml
@@ -8,6 +8,7 @@ defaultDefinition: {}
 discoverable: true
 category: DATA
 icon: select_check_box
+defaultVariant: uesio/io.default
 variants:
   - uesio/io.avatar:uesio/io.default
   - uesio/io.checkboxfield:uesio/io.default
--- a/libs/apps/uesio/io/bundle/components/group.yaml
+++ b/libs/apps/uesio/io/bundle/components/group.yaml
@@ -7,6 +7,7 @@ entrypoint: components/group/group
 discoverable: true
 variants:
   - uesio/io.group:uesio/io.default
+defaultVariant: uesio/io.default
 slots:
   - name: components
     label: Group Contents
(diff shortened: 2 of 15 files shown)


humandad:
I think this is good to merge. Are we ok with taking away the WIP?

humandad (approved):

Issue #4625, opened 2025-03-01
bug: `app delete` indicates successful delete for an app that does not exist


techfg:
When executing the `app delete` command, if the `name` provided for the app does not exist, the CLI returns `Successfully deleted app <appname>`.   

There is something to be said in treating this as a no-op since the user requests the app to be deleted and since it doesn't exist, it essentially is.  However, I think a more appropriate behavior would be to indicate "app does not exist" as a user may mis-type the input and think that the app was deleted when it really was not.

One thing to consider is having the CLI exit code be `0` (zero) to indicate success but with a different message but likely best to have exit code be non-zero along with message indicating `app not found` when the delete did not actually occur.

Steps to reproduce:
1) `uesio app delete -n thenameofanappthatdoesnotexist`
OR
2) `uesio app delete -n ben/thenameofanappthatdoesnotexist`

Actual Result
`Successfully deleted app thenameofanappthatdoesnotexist`
OR
`Successfully deleted app ben/thenameofanappthatdoesnotexist`

Expected Result
App `thenameofanappthatdoesnotexist` was not found with non-zero exit code
OR
App `ben/thenameofanappthatdoesnotexist` was not found with non-zero exit code

Related #4626

PR #4648, opened 2025-03-04
bug: app delete indicates successful delete for an app that does not exist

abeljimo:
# What does this PR do?

This fixes this issue https://github.com/ues-io/uesio/issues/4625 good catch @techfg, It's a load to the server every time but I guess this is our source of truth. 

# Testing

uesio app delete -n ben/thenameofanappthatdoesnotexist

**expected result** : Error: the requested app does not exist: ben/thenameofanappthatdoesnotexist


techfg:
@abeljimo -

Thanks for tackling this one!  I think the approach you took serves the purpose of #4625 regarding providing the user more meaningful output message.  That said, possibly an alternate approach to accomplishing would be to have the DELETE api return a 404 vs. it's current 204 when the app doesn't exist.  This would accomplish a couple of things:

1. Avoid the extra call to the server to check for existence
2. Keeps the API idempotent since the server state is the same after a DELETE on every attempt call but by having a 404, the client can handle the result differently

Thoughts?

techfg (commented):
@abeljimo -

Thanks for tackling this one!  I think the approach you took serves the purpose of #4625 regarding providing the user more meaningful output message.  That said, possibly an alternate approach to accomplishing would be to have the DELETE api return a 404 vs. it's current 204 when the app doesn't exist.  This would accomplish a couple of things:

1. Avoid the extra call to the server to check for existence
2. Keeps the API idempotent since the server state is the same after a DELETE on every attempt call but by having a 404, the client can handle the result differently

Thoughts?

abeljimo:
> @abeljimo -
> 
> Thanks for tackling this one! I think the approach you took serves the purpose of #4625 regarding providing the user more meaningful output message. That said, possibly an alternate approach to accomplishing would be to have the DELETE api return a 404 vs. it's current 204 when the app doesn't exist. This would accomplish a couple of things:
> 
> 1. Avoid the extra call to the server to check for existence
> 2. Keeps the API idempotent since the server state is the same after a DELETE on every attempt but by having a 404, the client can handle the result differently
> 
> Thoughts?

@techfg I've read a bit about it and your approach makes more sense, I'm working it out but I'll have to fix some tests and probably other things too! Thanks for the advice!

techfg:
> @techfg I've read a bit about it and your approach makes more sense, I'm working it out but I'll have to fix some tests and probably other things too! Thanks for the advice!

Sounds good @abeljimo!  Let me know if there's anything I can help with :)  One thing to note is that if we change this API to return 404, we should likely change all the other DELETE apis to behave the same.

Thanks!

abeljimo:
Hey @techfg that should be ready to review it makes sense to me. The other DELETE API already behave correctly.
We could use the ctlutil.HandleError function like you see in the image, but not sure if in the case we want to log every single delete to the server, I guess we don't. I hope everything is clear and thank you for the help!
 
Screenshot 2025-03-23 at 14 33 06

apps/cli/pkg/call/call.go
```diff
@@ -51,7 +51,7 @@ func Request(r *RequestSpec) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode >= 400 {
+	if resp.StatusCode >= 400 && resp.StatusCode != http.StatusNotFound {
```
techfg, line 54: This would change the behavior for every call that is made to Request throughout the entire CLI.  Previously, when http.StatusNotFound was encountered, the calling code would receive an error but with this change, it won't which means all the places we call Request would need to change to detect the StatusCode for a NotFound.  Additionally, those places would need to read the response body to obtain any information that was sent.  A more direct approach here would be to return different error types based on the status code.  For example:

```
	if resp.StatusCode >= 400 {
		defer resp.Body.Close()
		data, err := io.ReadAll(resp.Body)
		if err != nil {
			return nil, err
		}

		switch resp.StatusCode {
		case http.StatusNotFound:
			return nil, exceptions.NewNotFoundException("resource not found")
		case http.StatusForbidden, http.StatusUnauthorized:
			return nil, exceptions.NewForbiddenException("permission denied")
		default:
			return nil, fmt.Errorf("request to %v failed with status code %v: '%v'", resp.Request.URL, resp.StatusCode, string(data))
		}
	}
```

Also note that this code sample moves the resp.Body.Close to immediately after the StatusCode >= 400 check.  The existing code has a bug where if io.ReadAll fails, the response body would never be closed leading to connections being leaked.  Since this code path (statuscode > 400) is taking responsibility for the response and not returning it to the caller, it must be sure to always close the response body in all cases.

One last note is that the code below where we check locationHeader has the same resp.Body.Close() problem as it does not return the response but an error instead but without closing the response body.  This is a pre-existing bug but it can be fixed in this PR by adding resp.Body.Close before returning.  Unfortunately, GH doesn't allow commenting on a line that didn't change so I can't put the comment on that line so tracking it here instead :)

apps/cli/pkg/wire/save.go
```diff
@@ -43,16 +43,24 @@ func Insert(collectionName string, changes []map[string]interface{}, appContext
...
-	if statusCode, err := call.Delete(deleteUri, sessionId, appContext); err != nil || statusCode != http.StatusNoContent {
-		return false, err
+	statusCode, err := call.Delete(deleteUri, sessionId, appContext)
+	if err != nil {
+		return err
+	}
+	switch statusCode {
+	case http.StatusNoContent:
+		return nil
+	case http.StatusNotFound:
```
techfg, line 59: This case can be removed since call.Delete will now return a NotFoundException error

apps/platform-integration-tests/hurl_seeds/delete_app.hurl
```diff
@@ -11,11 +11,15 @@ user_id: jsonpath "$.user.id"
...
 x-uesio-app: uesio/tests
-HTTP 204
+HTTP *
+[Asserts]
+status toString matches /^(204|404)$/
 
 # Delete the tests app if it already exists, which should cascade delete the workspace and all of its metadata
 DELETE https://{{host}}:{{port}}/site/api/v1/collection/uesio/studio/app?uesio/studio.fullname=eq.uesio%2Ftests
-HTTP 204
+HTTP *
```
techfg, line 20: Due to the way our tests are written and executed, we may or may not have an app to delete so it makes sense to check for 204 or 404.  However, we should add tests in the hurl_specs folder to explicitly check deleting a app that is expected to be there returns a 204 (and not a 404) and then attempting to delete something that is not expected to be there returns a 404 and not a 204.  Easy approach would be to write a set of tests that do the following:

1. Create a new app
2. Delete the app asserting a 204
3. Delete the app again asserting a 404

Same set of tests for a workspace as well.

apps/cli/pkg/command/app/delete.go
```diff
@@ -28,13 +27,10 @@ func Delete(appFullName string) error {
 		}
 		appFullName = app.FullName
 	}
-
 	// Delete the app
-	success, err := wire.DeleteApp(appFullName)
+	err = wire.DeleteApp(appFullName)
 	if err != nil {
 		return fmt.Errorf("unable to delete app %s: %s", appFullName, err.Error())
```
techfg, line 33: This should change to make use of error wrapping via %w formatter to ensure the entire error is returned, not just the message

`fmt.Errorf("unable to delete app %s: %w", appFullName, err)

apps/cli/pkg/command/workspace/delete.go
```diff
@@ -33,13 +33,10 @@ func Delete(workspaceName string) error {
 	}
 
 	// Delete the workspace
-	success, err := wire.DeleteWorkspace(appObject.FullName, workspaceName)
+	err = wire.DeleteWorkspace(appObject.FullName, workspaceName)
 	if err != nil {
 		return fmt.Errorf("unable to delete workspace %s: %s", workspaceName, err.Error())
```
techfg, line 38: Wrap error with %w formatter
`fmt.Errorf("unable to delete workspace %s: %w", workspaceName, err)

techfg (changes requested):

techfg:
> Hey @techfg that should be ready to review it makes sense to me. The other DELETE API already behave correctly. We could use the ctlutil.HandleError function like you see in the image, but not sure if in the case we want to log every single delete to the server, I guess we don't. I hope everything is clear and thank you for the help!
> 
> Screenshot 2025-03-23 at 14 33 06

Hi @abeljimo, sorry for the delay in reviewing this!  I updated with some comments/feedback.  Agreed, we don't need to call `HandleError` which would trigger a log message on a 404, the way you have that coded currently works :)

Let me know if you have any questions on the feedback I left!

apps/cli/pkg/wire/save.go
abeljimo: I had to add this back since the call.Delete ends calling DeleteRecordApi and it returns also StatusNoContent when the element is deleted but nothing comes back in the response. without this I was getting an error check the previous commit. @techfg

techfg (changes requested):

techfg:
Looks great @abeljimo, thank you for doing this!!

Closes #4625

techfg (approved):

PR #4653, opened 2025-03-05
refactor default variant code

humandad:
# What does this PR do?

1. Simplifies the `parseVariantName` function - this function is no longer responsible for figuring out default variants if none exist. It now just determines a fully qualified variant key from a combination of variant key and component type. It is the responsibility of code that calls `parseVariantName` to use a default if the key provided is not valid/present.

2. Moves the `getVariantDefinition` function from `styles.tsx` to `component.tsx`. This function is now responsible for loading a default variant from the component definition if one is provided.

3. Simplifies the `context.getComponentVariant` function. It now is only responsible for getting a valid component variant from the redux store. It does not process/parse the name provided and just uses it "as is".


techfg (commented):
Much more straightforward approach!  Tested several scenarios and all look good and didn't see any console logs emitted.  Left one other comment but otherwise, this should be good to go!

PR #4654, opened 2025-03-05
Upgrade monaco and turn off sticky scroll

humandad:
# What does this PR do?

This is one option for upgrading monaco to 0.52.2. If I turn off the sticky scroll option, I no longer get the test failures I was experiencing before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,15 +613,10 @@ For Go **package naming**, we follow this [guideline](https://blog.golang.org/pa
 
 ## Managing Dependencies
 
-1. We're pinning monaco to version 0.50.0 for now because of this [bug](https://github.com/microsoft/monaco-editor/issues/4654).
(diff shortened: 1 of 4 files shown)


techfg (commented):
Interesting that this avoids triggering the underlying issue.

My $0.02 is that we stick with `0.50.0` until monaco-editor releases a fix or we find a solution that doesn't require disabling features we could/currently are using.  We also can't be 100% certain the issue won't crop up somewhere even with sticky disabled.

If you do want to merge this, only thing I'd suggest adding is a TODO comment [here](https://github.com/ues-io/uesio/pull/4654/files#diff-66563c3d3626b5d21fcb5fb31e1979f2aa36b4315bd3223824a6896a4b27cfaaR175) that references the monaco-editor bug we've been tracking so we can revert back to using sticky once it's addressed.

humandad:
Yeah let's just keep this out for now. I need to stop thinking about monaco for at least a month!

PR #4655, opened 2025-03-05
Make code panel height adjustable

humandad:
# What does this PR do?

Allows for adjusting the code panel height in the builder.


libs/apps/uesio/builder/bundle/componentpacks/main/src/utilities/adjustableheightarea/adjustableheightarea.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const handleMouseMove = (e: MouseEvent) => {
+    e.preventDefault()
+    const newHeight = startHeight.current + (startY.current - e.clientY)
+    setHeight(newHeight > 0 ? newHeight : 0)
+  }
+
+  const handleMouseUp = (e: MouseEvent) => {
+    e.preventDefault()
+    document.removeEventListener("mousemove", handleMouseMove)
```
techfg, line 45: Any reason not to use `AbortController` instead of `removeEventListener`?

techfg (commented):
Overall, looks good, great UX improvement!

The one thing we don't handle is keyboard accessiblity for the height but we have a general keyboard accessibility issue throughout the platform currently so I think it can wait until we focus on accessibility in general. 

Left one other comment but other than that potential change, I think this is good to go, thank you!

libs/apps/uesio/builder/bundle/componentpacks/main/src/utilities/adjustableheightarea/adjustableheightarea.tsx
humandad: Hmmm. I've never even heard of AbortController before. From the docs it seems like the main use case is for the fetch api, not as an alternative to removeEventListener. But maybe I just don't know enough about AbortController yet. Maybe we could do a followup PR if we had a good reason to adjust?

techfg: Yeah, AbortController's roots were in cancelling things like fetch, etc. but it expanded in to event listeners via the [signal`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal) option passed in to `addEventListener`.  It's a greatly simplified way to cancel listeners vs. having to track the function and then call `removeEventListener` especially helpful when you have several listeners.  Something to use moving forward I think and we can always go back and adjust existing add/removeEventListener code in separate PR if/when desired.

PR #4678, opened 2025-03-13
Fix: Defaults should override explicit undefined definition

humandad:
# What does this PR do?

This fixes our implementation of defaultValues for props and defaultContent for slots. Before we were spreading the defaults before the definition. Now we spread them after the definition.

This fixes the case where the definition could have included an explicit undefined value like

```
{
  myprop: undefined
}
```

Although a default would be generated for this property, the spread operation would actually override it back to the undefined value.

# Testing

Added a failing test for this case.

### DIFF
--- a/libs/ui/src/component/component.spec.tsx
+++ b/libs/ui/src/component/component.spec.tsx
@@ -364,6 +364,23 @@ const addDefaultPropertyAndSlotValuesTests = [
       title: "We provided our own title",
     },
   },
+  {
+    name: "value provided for props but not slots, however, the props are explicitly set to undefined",
+    inputDefinition: {
+      title: undefined,
+    },
+    componentDef: componentTypeWithSlotAndPropertyDefaults,
+    expected: {
+      header: [
+        {
+          "uesio/io.titlebar": {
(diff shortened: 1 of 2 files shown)


techfg (approved):
I think this change makes sense since the underlying component is the lowest level so it's defaults should apply when the prop value is `undefined` - it's akin to passing undefined to a function parameter and having the default of the function parameter value be used.

The only two potential risks that I can think of in this approach are:
1. If a component really wants something to be 'undefined' and not use the underlying `defaultValue` it would not be possible
2. If properties/defaultValue's can be complex objects and if a sub-properties property has a default value, that could be lost if the definition defined the top-level property but didn't contain a value for the sub-property property since spread syntax is a shallow merge instead of deep.  If this scenario is even one we support (i'm not sure), this is likely OK/acceptable as long as we document this case clearly.

There really isn't an ideal solution for these types of scenarios for which there are several in the platform where we layer in defaults.  The main reasons being:

1. Explicit use of `undefined` intentionally to avoid a defaultValue
2. Shallow vs deep merge handling

I think the key is to be consistent in each scenario across the board (which I believe this PR gets us closer to) and also document it clearly (which we will need to do).

PR #4680, opened 2025-03-13
chore: use appkit detail form for the file detail page

humandad:
# What does this PR do?

Switches to using the more standardized appkit form detail component for the file detail page. (we should switch other detail pages to use this as well)

### DIFF
--- a/libs/apps/uesio/appkit/bundle/components/form_detail.yaml
+++ b/libs/apps/uesio/appkit/bundle/components/form_detail.yaml
@@ -142,10 +142,13 @@ definition:
                     - type: hasNoValue
                       value: $Prop{initials}
                   text: $If{[$Prop{avataricon}][$Prop{avataricon}][$Collection{icon}]}
+                  color: $Prop{avatariconcolor}
                   uesio.variant: uesio/io.icon
         - $Slot{content}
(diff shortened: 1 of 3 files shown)


techfg:
@humandad -

Tested this, a few issues/questions:

1. The files dropzone is active in "READ" mode - this is the way it was before but with the READ/EDIT concept now, should it be required to go in to "EDIT" mode in order to add/remove individual files?
2. In either READ/EDIT mode, you can click the dropzone inside of an existing file already there to add another file but this seems a little odd since you are clicking inside of an existing file to add another file vs. an "open" area drop-zone to add another file
3. Just confirming that a "uesio File" can actually contain multiple physical files, correct?
4. When clicking "DELETE" from "EDIT" mode, the dialog displayed is empty when there no physical files or when there are physical files
5. On narrower viewport, there are some styling/CSS issues which seems to trigger more often on files with longer names
image
6. For files with longer names or even a shorter name but a narrower viewport the file size text area become garbled 
image

humandad:
I'll answer the others later, but
3. Yes
4. This is actually fixed by merging in the default values fix that I submitted earlier today.

techfg:
sounds good! testing that PR now, thanks! :)

PR #4681, opened 2025-03-14
Remove id prop from base definition

humandad:
# What does this PR do?

Removes a backwards-compat hack we put in a while ago to support the Morandi app. I just went through the morandi app and fixed all this, so I think we can go ahead and remove the hack.

We no longer support the `id` property on uesio components. You have to use `uesio.id` to provide the id.


humandad:
I also had to fix a bug in `propertiesform.tsx` to allow the type checker to pass.

PR #4696, opened 2025-03-17
Add agent metadata type

humandad:
# What does this PR do?

1. Adds the `agent` metadata type.
2. Adds the `manage_agents` feature flag.
3. Adds basic studio UI for managing agents.

Screenshot 2025-03-17 at 10 11 27 AM

NOTE:

This PR does not contain any actual agent functionality, just the ability to create/deploy/retrieve agent metadata items. That's why it's all behind a feature flag.

# Testing

1. Turn on the `manage_agents` feature flag.
2. Basic UI and deploy/retrieve should work for agents.

### DIFF
--- a/apps/platform/pkg/deploy/deploy.go
+++ b/apps/platform/pkg/deploy/deploy.go
@@ -54,6 +54,7 @@ var ORDERED_ITEMS = [...]string{
 	"integrationtypes",
 	"integrations",
 	"integrationactions",
+	"agents",
 }
 
 type DeployOptions struct {
--- a/apps/platform/pkg/meta/agent.go
+++ b/apps/platform/pkg/meta/agent.go
@@ -22,7 +22,7 @@ func NewBaseAgent(namespace, name string) *Agent {
 type Agent struct {
 	BuiltIn        `yaml:",inline"`
 	BundleableBase `yaml:",inline"`
-	Description    string `yaml:"description" json:"description"`
+	Description    string `yaml:"description" json:"uesio/studio.description"`
 	ProfileRef     string `yaml:"profile" json:"uesio/studio.profile"`
 }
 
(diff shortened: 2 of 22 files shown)


techfg (commented):
All looks good.  Couple of thoughts/considerations:

1. When viewing an "agent", the "prompt" is displayed in a monaco editor - if this is only ever intended to display the prompt itself, possibly this should just be a text area instead?  If remains an editor, it's just text so assuming yaml is currently being set for language or?
2. There is no label for the "prompt" field above the editor area.  This is true on all "views" of this type actually but for things like a "view" itself, the definition of the "view" is in the editor so not having a label kind of makes sense but possibly for things like this, the "field" on the screen should have a label (e.g., System Prompt)?
3. Not sure the final plan but can a prompt be "edited" once the bot is doing it's thing?  If not, then we'll need to think about how to not allowing "edit" if you go in to edit mode.

humandad:
1. The language is currently set for plaintext. I think monaco editor is fine here since this prompt could possibly be quite long. LLMs like claude understand XML and Markdown as prompts (or any language for that matter), so I guess we could later allow for prompt.md or prompt.xml. I think using monaco editor with the language set to "plaintext" is fine here for now.

Check out this version of v0's prompt.
https://github.com/sharkqwy/v0prompt/blob/main/prompt.txt

2. Agreed. I'll add a label.
3. Editing a prompt is fine here. This is metadata. So any change to metadata will have to be part of a bundle upgrade for a site. Editing this prompt is editing it in workspace mode so any new agent conversation action will use the new prompt. It is a good question about whether an ongoing conversation with an agent will switch to the new system prompt after an upgrade or continue to use the old one. (But that is more an implementation detail) not related to the metadata editing experience.

techfg (approved):

PR #4722, opened 2025-03-25
feature: Add Claude to supported models list

humandad:
# What does this PR do?

1. Adds support for the `anthropic.claude-3-5-sonnet-20241022-v2:0` model in our bedrock integration.
2. Improves the `GetClientOptions` function so that it can receive the input parameters before generating an options function.


apps/platform/pkg/integ/bedrock/claude.go
```diff
@@ -48,7 +65,17 @@ type ClaudeModelHandler struct {
 
 var claudeModelHandler = &ClaudeModelHandler{}
 
-func (cmh *ClaudeModelHandler) GetClientOptions(o *bedrockruntime.Options) {}
+func (cmh *ClaudeModelHandler) GetClientOptions(input *bedrockruntime.InvokeModelInput) func(o *bedrockruntime.Options) {
+
+	// Force claude 3.5 v2 to use the us west model
```
techfg, line 70: Assuming this is temporary?  Should it be annotated with a TODO so we don't lose sight of it and possibly a const since we use in multiple places?

apps/platform/pkg/integ/bedrock/claude.go
```diff
@@ -64,9 +91,16 @@ func (cmh *ClaudeModelHandler) GetBody(options *InvokeModelOptions) ([]byte, err
 		})
 	}
 
+	anthropicBeta := []string{}
+	// Add the antropic beta flag
+	if options.Model == "anthropic.claude-3-5-sonnet-20241022-v2:0" {
```
techfg, line 96: Same question from above re: TODO and const :)

apps/platform/pkg/integ/bedrock/stability.go
```diff
@@ -23,8 +23,11 @@ type StabilityModelHandler struct {
 
 var stabilityModelHandler = &StabilityModelHandler{}
 
-func (smh *StabilityModelHandler) GetClientOptions(o *bedrockruntime.Options) {
-	o.Region = "us-west-2"
+func (smh *StabilityModelHandler) GetClientOptions(input *bedrockruntime.InvokeModelInput) func(o *bedrockruntime.Options) {
+	return func(o *bedrockruntime.Options) {
+		// The stability models only work in the us-west-2 region
+		o.Region = "us-west-2"
```
techfg, line 29: same question as above re: TODO and const :)

PR #4729, opened 2025-03-25
Refactor: Separate route to handle dynamic view loads

humandad:
# What does this PR do?

Previously, dynamically loaded views would run the `getBuilderDeps` function. This function would pull in the metadata needed to load a view, but would also load in a lot of unnecessary things that the builder needs. This should be a runtime function that works for all dynamically loaded views.


libs/ui/src/components/view.tsx
```diff
@@ -101,7 +101,7 @@ const View: UC<ViewComponentDefinition> = (props) => {
     if (viewDefId && !viewDef) {
       memoizedAsync(
```
techfg, line 102: Using memorized here, if the view changes will this ensure the latest gets loaded?

libs/ui/src/hooks/builderapi.ts
```diff
@@ -49,4 +49,19 @@ const getBuilderDeps = async (context: Context) => {
...
 
-export { useMetadataList, useAvailableNamespaces, getBuilderDeps }
+const loadBuilderDeps = async (context: Context) => {
+  const response = await platform.getViewDeps(context)
+
+  await loadScripts(response.componentpack, context)
+
+  dispatchRouteDeps(response)
+
+  return
```
techfg, line 59: can remove return statement

libs/ui/src/components/view.tsx
humandad: No, but I think that's a tradeoff that is ok. We will use the cached version until the user refreshes the page.

PR #4736, opened 2025-03-26
feat: support integration site merges

techfg:
# What does this PR do?

Adds support for `$Site` merges on GetBaseURL to integration bots.  Previously, $Site was only supported in systemdialect bots.

Note - One other non-breaking change is that the previous code was intending to "hide/abstract" the `Integration` type from non-system dialects via IntegrationMetadata using a type declaration.  This was hiding/abstracting `Integration` but not the embedded types in `Integration` (e.g., BuitlIn, BundleBase, etc.).  This PR avoids type embedding for `IntegrationMetadata` ensuring that only what we want to expose is exposed.

# Testing

ci & e2e will cover.


humandad (approved):

humandad:
This is great! Thanks for doing this, Barry.

PR #4743, opened 2025-03-28
feat: add `$Site{scheme}` merge option

techfg:
# What does this PR do?

Adds $Site{scheme} context merge to server & client.  This will help support building urls in yaml, tests, etc. and specifically is a step towards being able to run the platform locally with SSL or without SSL (e.g., eventually being able to simply use localhost instead of studio.uesio-dev.com).

@humandad - Beyond general review, of specific note is the change to the oauth integration yaml for tests.  I could not find anywhere where `baseUrl` was used for these so I removed them completely.  It looks like we use `authorizeUrl` from the credentials.yaml but do not use `baseUrl` anywhere.  For example, see the [salesforce integration](https://github.com/ues-io/integrations/blob/main/apps/salesforce/bundle/integrations/salesforce.yaml) and [credentials](https://github.com/ues-io/integrations/blob/main/apps/salesforce/bundle/credentials/salesforce.yaml).  Let me know if you think/find otherwise.

# Testing

Manual testing confirms all works in client & server, ci & e2e will cover rest.

### DIFF
--- a/apps/cli/pkg/call/call.go
+++ b/apps/cli/pkg/call/call.go
@@ -12,6 +12,7 @@ import (
 	"github.com/thecloudmasters/cli/pkg/config/host"
 	"github.com/thecloudmasters/cli/pkg/context"
 	"github.com/thecloudmasters/uesio/pkg/controller/ctlutil"
+	"github.com/thecloudmasters/uesio/pkg/types/exceptions"
 )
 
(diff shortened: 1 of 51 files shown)


humandad (approved):
Looks good! Thanks for doing this. Removing baseUrl from the oauth tests should be fine.

PR #4748, opened 2025-03-28
Improve comment thread stying and data model

humandad:
# What does this PR do?

1. Adds the `author` field to the `thead_item` collection. This will now be used to differentiate between USER and ASSISTANT items.
2. Changes the meaning of the `type` field on the `thread_item` collection. This will no be used to specify the type of item (text/tool_use/tool_result)
3. Adds `tool_input`, `tool_use_id` and `tool_name` fields to `thread_item`.
4. Add `agent_user` permission set to aikit for simplifying permissions.
5. Revamp/Refactor thread_item component styling and rendering.

NOTE:

This will cause old AI threads in the crm app to have their content not show up. But I don't think anyone is actually using the thread data there since it's basically useless (just inference on a model). New threads will work just fine. (It's mostly used for demo purposes anyways).


libs/apps/uesio/aikit/bundle/bots/listener/ai_chat/bot.ts
```diff
@@ -28,7 +31,7 @@ export default function ai_chat(bot: ListenerBotApi) {
   // Loop over the messages and put them in the right format
   const messages = messagesResult
     .map((message) => ({
-      role: message["uesio/aikit.type"] === "USER" ? "user" : "assistant",
+      role: message["uesio/aikit.author"] === "USER" ? "user" : "assistant",
```
techfg, line 34: Should these values (USER/user/assistant) become constants since they are used in several places?

techfg (approved):

Issue #4752, opened 2025-03-29
test: e2e test `create a new view` in `builder.cy.ts` fails due to `ResizeObserver loop completed with undelivered notifications` errors


techfg:
When running [Create a new View](https://github.com/ues-io/uesio/blob/main/apps/platform-e2e/cypress/e2e/builder.cy.ts#L160) test, the test fails with the following error:

```
     Error: The following error originated from your application code, not from Cypress.

  > ResizeObserver loop completed with undelivered notifications.

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application
    at <unknown> (https://studio.uesio-dev.com:3000/workspace/uesio/e2eenwepziz/test/views/uesio/e2eenwepziz/testview/edit:0:0)
```

A few notes:
1. This test was not failing when run locally prior to 0fdb2476f6fbb8e58b3dac7804ff319baf5bc894
2. It appears that this test is [skipped on CI]() as of 63c83343953b63aec071a4eb13ff8dfbd3996b47 - See #4751.  It is unknown if the issue seen now locally is the same as what was occuring on CI or they are different since there aren't any details on why these were disabled in CI (e.g., error messages, etc.).
3. The problem does not occur if `display: grid` from [djustableHeightArea](https://github.com/ues-io/uesio/blob/main/libs/apps/uesio/builder/bundle/componentpacks/main/src/utilities/adjustableheightarea/adjustableheightarea.tsx#L58) is removed.  This property was added in 0fdb2476f6fbb8e58b3dac7804ff319baf5bc894.  By itself, it should not create any issue but the extra layout cycles are likely triggering the underlying issue.  We only use a ResizeObserver in [tablabels.tsx](https://github.com/ues-io/uesio/blob/main/libs/apps/uesio/io/bundle/componentpacks/main/src/utilities/tablabels/tablabels.tsx#L5) so likely is one of our dependencies and/or the just a byproduct of browser behavior running under cypress test.
4. The problem occurs when we send [{meta}y](https://github.com/ues-io/uesio/blob/main/apps/platform-e2e/cypress/e2e/builder.cy.ts#L186) which opens the code panel triggering the layout to adjust.

The underlying issue requires further investigation to determine its root source, either in our own code and/or in a dependency although the issue could just be the way the browser performs layouts and running under Cypress control.

[...]

PR #4754, opened 2025-03-29
fix local builder.cy test

humandad:
# What does this PR do?

This should at least allow us to run builder.cy locally now. I think adding the "grid" css style to the wrapper around the code panel was causing it to animate the monaco editor container. That was probably firing a lot of resize events and possibly making thing crazy in Cypress.

This change fixes the scrolling issue (which I was originally trying to fix) without causing the monaco editor inside the panel to animate its height.

I would test this locally, but for some reason I can't get cypress to run locally anymore.


techfg:
@humandad - Thanks for this and makes sense!  I tested your fix locally with cypress and it resolves the issue.  I'll circle back to this with a little more testing and then merge once I finish up current work items :)

techfg (approved):
Looks great, thanks for taking care of this @humandad!

techfg:
Closes #4752

PR #4765, opened 2025-03-31
chore: `UESIO_` prefix for all environment variables

techfg:
# What does this PR do?

Modifies all UESIO environment variables to have a `UESIO_` prefix.

Requires coordination with uesio-infra deployment to modify env vars.

# Testing

ci & e2e will cover.


humandad (approved):
This is great. I love that we're getting some more consistency here.

PR #4777, opened 2025-04-03
Cleanup bad request exception type

humandad:
# What does this PR do?

1. Changes the NewBadRequestException constructor to take an error instead of a string.


apps/platform/pkg/adapt/postgresio/errors.go
```diff
@@ -16,13 +16,15 @@ func TranslatePGError(err error) error {
 		switch pgErr.Code {
 		case "22P02":
 			// invalid input syntax
-			return exceptions.NewBadRequestException(pgErr.Message)
+			// We're intentionally hiding the details of the error
+			// for example: pgErr.Severity and pgErr.Code
+			return exceptions.NewBadRequestException(pgErr.Message, nil)
```
techfg, line 21: Should we slog here to capture full error?

apps/platform/pkg/adapt/postgresio/errors.go
```diff
@@ -16,13 +16,15 @@ func TranslatePGError(err error) error {
...
 		}
 	}
 
 	msg := err.Error()
 	if strings.Contains(msg, "failed to encode") || strings.Contains(msg, "failed to decode") {
-		return exceptions.NewBadRequestException(msg)
+		return exceptions.NewBadRequestException(msg, nil)
```
techfg, line 27: slog to capture full error since we're dropping it?

apps/platform/pkg/auth/google/googleauth.go
```diff
@@ -143,10 +143,10 @@ func (c *Connection) Signup(signupMethod *meta.SignupMethod, payload map[string]
 	return c.callListenerBot(signupMethod.SignupBot, payload)
 }
 func (c *Connection) ResetPassword(payload map[string]interface{}, authenticated bool) (*meta.LoginMethod, error) {
-	return nil, exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password")
+	return nil, exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password", nil)
```
techfg, line 146: should this be a 403 Forbidden instead of 400?

apps/platform/pkg/auth/google/googleauth.go
```diff
@@ -143,10 +143,10 @@ func (c *Connection) Signup(signupMethod *meta.SignupMethod, payload map[string]
...
 }
 func (c *Connection) ConfirmResetPassword(payload map[string]interface{}) (*meta.User, error) {
-	return nil, exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password")
+	return nil, exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password", nil)
```
techfg, line 149: should this be a 403 Forbidden instead of 400?

apps/platform/pkg/auth/google/googleauth.go
```diff
@@ -161,5 +161,5 @@ func (c *Connection) CreateLogin(signupMethod *meta.SignupMethod, payload map[st
 	}, c.connection, c.session)
 }
 func (c *Connection) ConfirmSignUp(signupMethod *meta.SignupMethod, payload map[string]interface{}) error {
-	return exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password")
+	return exceptions.NewBadRequestException("Google login: unfortunately you cannot change the password", nil)
```
techfg, line 164: should this be a 403 Forbidden instead of 400?

apps/platform/pkg/auth/mock/mockauth.go
```diff
@@ -41,7 +41,7 @@ func (c *Connection) Login(w http.ResponseWriter, r *http.Request) {
 	var loginRequest map[string]interface{}
 	err := json.NewDecoder(r.Body).Decode(&loginRequest)
 	if err != nil {
-		ctlutil.HandleError(w, exceptions.NewBadRequestException("invalid login request body"))
+		ctlutil.HandleError(w, exceptions.NewBadRequestException("invalid login request body", nil))
 		return
```
techfg, line 45: slog since we're dropping origina error?

apps/platform/pkg/auth/platform/platform.go
```diff
@@ -105,7 +105,7 @@ func (c *Connection) Login(w http.ResponseWriter, r *http.Request) {
 	var loginRequest map[string]interface{}
 	err := json.NewDecoder(r.Body).Decode(&loginRequest)
 	if err != nil {
-		ctlutil.HandleError(w, exceptions.NewBadRequestException("invalid login request body"))
+		ctlutil.HandleError(w, exceptions.NewBadRequestException("invalid login request body", nil))
```
techfg, line 108: slog since we're dropping original error?

apps/platform/pkg/datasource/load.go
```diff
@@ -73,10 +73,6 @@ func LoadWithError(op *wire.LoadOp, session *sess.Session, options *LoadOptions)
 	}
 
 	if op.Errors != nil {
-		// TODO: Using [0] results in losing the original error in some cases which results in not always returning
-		// an appropriate HTTP status code.  For example, the following url which has an invalid format for userfileid
-		// param encounters HTTP 500 when it should be HTTP 400
-		// https://studio.uesio-dev.com:3000/site/userfiles/download?userfileid=my-bad-id&version=1734112100
```
techfg, line 79: We should still keep this comment & TODO I think since we're still returning `(*op.Errors)[0]` which won't always have the original error

apps/platform/pkg/datasource/save.go
```diff
@@ -170,9 +171,13 @@ func SaveOp(op *wire.SaveOp, connection wire.Connection, session *sess.Session)
 		return HandleErrorAndAddToSaveOp(op, err)
 	}
 
-	// Check for population errors here
+	// NOTE: for actual errors with the population function,
+	// Populate desn't return an error, but it does add the errors
+	// to the op. We check for those errors here.
 	if op.HasErrors() {
-		return wire.NewGenericSaveException(exceptions.NewBadRequestException("error with field population"))
+		// TODO: This error message should be better and possibly include messages
```
techfg, line 178: Should likely slog all the errors in op here since we're dropping them

apps/platform/pkg/types/exceptions/badrequest.go
```diff
@@ -1,12 +1,15 @@
 package exceptions
 
-type BadRequestException struct {
-	msg string
-}
+type BadRequestException BaseException
```
techfg, line 3: I don't think we gain anything (other than the private vars due to same memory layout) by the type declaration here since BadRequestException won't get BaseExceptions methods and we're customizing them anyway.  I think the options here are:

1. Embed BaseException - Not sure if this should be the type `BaseException` or pointer to type `*BaseException`, would need to test
```
type BadRequestException struct {
  BaseException
```

2. Just use the private package helpers w/out involving BaseException here.  

I think there will be times that we have exceptions that just need the BaseException functionality so I'd lean towards option 1 so that we define custom exceptions all the same way, whether or not there is any custom `Error` function or just using the default provided by `BaseException`

apps/platform/pkg/types/exceptions/load.go
```diff
@@ -1,35 +1,21 @@
...
+	// We put the formatted message in here so that
+	// Marshalling works out of the box.
+	// The other option is a custom marshaller
+	return &LoadException{printErr("", message, err), err}
 }
 
 type LoadException struct {
 	Message string `json:"message"`
-	error   error
+	err     error
```
techfg, line 12: See comment for BadRequestException about embedding, vs not, vs type declaration vs just using privates.  This one we expose `Message` so that it serializes - that requirement may change the general approach to having a baseexception.  I think the key is that we want to try to have all our custom errors follow the same definition approach if possible.

apps/platform/pkg/types/exceptions/save.go
```diff
@@ -1,41 +1,28 @@
...
+		err:      err,
 	}
 }
 
 type SaveException struct {
 	RecordID string `json:"recordid"`
 	FieldID  string `json:"fieldid"`
 	Message  string `json:"message"`
-	error    error
+	err      error
```
techfg, line 19: See comments on BadRequestException and LoadException about how to define the custom errors.

techfg (commented):
@humandad -

Left some comments but didn't annotate every situations since there were a lot of the same question/suggestion/comment.

1. I think anywhere we do not return the original error, we should make sure to slog it prior to returning.  We want to have the actual error somewhere.
2. There's a few places we BadRequestException that possibly should be a different exception type
3. Some questions on how we define our custom error types - I think the key here is to be consistent across all of them (assuming that's possible)

PR #4780, opened 2025-04-05
chore: capture ci docker logs as artifact

techfg:
# What does this PR do?

1. Captures docker logs during CI as workflow artifact
2. Fixes the issue where we could infinitely wait for the app to start during health check by migrating to using native docker health checks
3. Adds an input parameter to manual CI runs (`workflow_dispatch`) to control the `UESIO_LOG_LEVEL` variable.  The default `UESIO_LOG_LEVEL` during CI is `info` (0) but when doing manual runs, this can be set to any level to aid with troubleshooting.

# Testing

ci will cover.


humandad (approved):
This is cool!

Issue #4781, opened 2025-04-05
bug: use of http clients does not adequately consider a multi-tenant scenario and is exposed to starvation


techfg:
Currently, most of the code in the CLI and platform uses [http.DefaultClient](https://pkg.go.dev/net/http#pkg-variables) except for the oauth2 package which [creates its own](https://github.com/ues-io/uesio/blob/main/apps/platform/pkg/oauth2/client.go#L183).

There are several issues with the current approach:

1. DefaultClient has a default [Timeout](https://pkg.go.dev/net/http#Client.Timeout) of zero.  This could lead to requests "hanging" indefinitely and eventually starving the available resources on the server.
2. DefaultTransport (which DefaultClient uses if Transport is not explicitly set) has a [MaxIdleConns](https://pkg.go.dev/net/http#DefaultTransport) of 100.  This setting is shared across all hosts meaning that if one bot/job/etc. hitting endpoint `some.server.com` is more active than other bot/jobs hitting other endpoints, `some.server.com` could make up most of the idle connections available.  This is even more impactful in a multi-tenant scenario because the http client is shared across all requests for the entire server currently.  There is a setting [MaxIdleConnsPerHost](https://pkg.go.dev/net/http#Transport.MaxIdleConnsPerHost) but the default for this is zero - this could be tuned but the heuristics to find the right value for it as a single point of control would not be trivial in a multi-tenant scenario.
3. There is currently no explicit consideration given to handle multi-tenant scenario in general
4. There is currently no ability for any bot/job/etc. to provide overrides to things like Timeout.  For example, if it's known that a certain request can take several minutes, this should at least be configurable as needed.  Currently, since the value is zero, it's not an issue for timeout but zero is not a reasonable default given issue # 1 above.

In short, how http clients are used internally within the core platform itself and within bot/jobs/etc. that may be uesio and/or third-party based needs to be evaluated and an approach defined that accounts for the issues mentioned above along with any others that can be thought of as the solution & architecture are considered.

PR #4782, opened 2025-04-06
chore/fix: resolve localhost domains to loopback address in go code & set http client timeout to 30 seconds

techfg:
# What does this PR do?

1. Implements a custom `http.Transport` that explicitly resolves `localhost` and `*.localhost` to the machines loopback address.  Golang stdlib makes a DNS call for any `localhost` domain which is different than browsers, curl, etc. which explicitly assign loopback address for security reasons.  This change is primarily targeting development of the platform to avoid the requirement of having to setup a `/etc/hosts` entry for the localhost subdomains when issuing golang http client.
2. Fixes an issue where the previous code was using the `http.DefaultClient` which has a default `Timeout` of zero which could lead to server starvation (see #4781)
3. Updates the scripts that generate SSL certificates and add local DNS entries to be more flexible, accept arguments, etc.  This change makes it much easier to run the platform from a custom domain on your localhost if that is needed.  By default, with the changes in this PR, everything will work with studio.localhost over HTTP or HTTPS, the only requirement is generating the SSL certificate if HTTPS is needed.
4. Adds a `Uesio Primary Domain` input argument to manual CI workflow runs in order to support running tests against a domain other than `localhost`.

> [!NOTE]
> The previous code from [this commit](https://github.com/ues-io/uesio/commit/b8f47528c3876af49bb9b37d0b4a2ea7bf623800#diff-9ffc606169aa6bc32d2e7367e2aa5a956a7dd4c5e0ff51624e21819e5b8ff57dR34) had logic that would add the SSL certificate to the TLS config of the `DefaultTransport` used by the http client.  It is not 100% clear why this code added since the golang client handles self-signed certs properly.  Possibly back when it was added golang did not handle self-signed certs.  What's odd though is that the CLI used `DefaultClient` directly, without the TLS configuration, and it made requests to the same local server running HTTPS.  Either way, this logic has been removed but making a note of it in case it causes a regression.

@humandad - Beyond general review, please make sure to test the ssl certification generation and local dns entries for custom domains on macOS.  See the changes to the README and details there for information on [custom domains setup](https://github.com/ues-io/uesio/pull/4782/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R206).

# Testing

Manual testing confirms all expected behavior - no `hosts` entries required anymore for `localhost` access to the site, CLI execution or tests.  ci & e2e will cover the rest.


techfg:
Update - I was able to figure out why the SSL cert was applied to TLS config in platform but wasn't in CLI.  In short, in local dev, we create a self-signed cert and as part of that process, add the generated cert to the trusted certificate store.  However, in CI, we do that on the "runner" machine, not the docker image.  We mount the certs to `/ssl` directory on the container so that the platform can load the cert, however its not copied to the container itself during build nor is it mounted to on run.

We don't want to copy the cert to the image because then all of our images would have it so until we potentially separate our build/test/release image orchestration, we need an alternative way to get the cert trusted inside of the docker container itself because as part of the e2e & integration tests, the server makes HTTP calls (to itself but still an http call - e.g., weather forecast "external" api tests).

One option is to mount the GH runners certs to the container (e.g.,  `/usr/local/share/ca-certificates:/usr/local/share/ca-certificates:ro`) and then run `update-ca-certificates` in the container command.  This works, but unfortunately `update-ca-certificates` does not ship with `alpine:latest` so we need to apk it in dockerfile (e.g., `RUN apk add --upgrade --no-cache ca-certificates && update-ca-certificates`).  This isn't terrible and works but the only reason we need this additional size and package in the base uesio image is for tests.

An alternate option is to read the cert in the server like the previous code did but that's not really ideal either.  At the very least, if we do do that, it should only be for CI runs (or worst case, also local dev).

More to come...

techfg:
Pushed a commit that includes adding the `ca-certificates` apk and then copying the generated cert and updating certs and all works.

There are pros/cons to reading our own cert and applying to TLS config vs. the apk approach.  The one advantage of having the apk in the base image is that anyone using the uesio image that wants to use their own cert (a self-signed one or more likely their own cert), the tooling will already be in the image.  Of course, in that scenario, they could compose an image based on our image that then contains the apk so it's trivial to solve for since alpine doesn't include `update-ca-certificates` by default.

The image size difference is negligible so that's not really a concern.  It comes down to installing the apk or going back to reading the cert we use for serving the platform anyway and applying it to http client tls config calls so that we trust ourselves.

Open to input on this, not a strong lean either way currently.

PR #4826, opened 2025-04-15
refactor: remove redundant loginmethod query

humandad:
# What does this PR do?

This issue was caught in a golangci-lint check for unused variables.

`loginmethod` was queried and never used until it was overwritten with another query. I can't think of any reason why we'd need to query this twice.

### DIFF
--- a/apps/platform/pkg/auth/platform/platform.go
+++ b/apps/platform/pkg/auth/platform/platform.go
@@ -248,20 +248,12 @@ func (c *Connection) ResetPassword(payload map[string]interface{}, authenticated
 		return nil, exceptions.NewBadRequestException("Unable to reset password: you must provide a username", nil)
 	}
 
-	code := generateCode()
-
-	adminSession := sess.GetAnonSessionFrom(c.session)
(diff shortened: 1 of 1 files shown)


apps/platform/pkg/auth/platform/platform.go
```diff
@@ -248,27 +248,22 @@ func (c *Connection) ResetPassword(payload map[string]interface{}, authenticated
...
+	loginmethod, err := auth.GetLoginMethodByUserID(user.ID, c.authSource.GetKey(), c.connection, c.session)
 	if err != nil {
 		return nil, err
 	}
 
 	if loginmethod == nil {
 		return nil, exceptions.NewBadRequestException("No account found with this login method", nil)
 	}
+
+	code := generateCode()
```
techfg, line 265: Unless I'm overlooking something, could eliminate the local var `code` here and just inline the call to generateCode on line 268.  We don't seem to use `code` anywhere else.

techfg: disregard

techfg (approved):

PR #4830, opened 2025-04-16
fix: respect workspace in collection & integration type after save bots

### LINKED ISSUE #4801: bug: deleting workspace "A" effects other workspaces
When deleting a workspace, it has the potential to effect other workspaces:

**Steps to reproduce**
1. Create new app
2. Create "dev" workspace
3. Add collection "Product" with a field "name"
4. Create "test" workspace
5. Add collection "Product" with a field "name"
6. Delete "test" workspace

**Actual Result**
The "Product" collection in "dev" workspace no longer has any fields

**Expected Result**
The "dev" workspace should be unaffected

Need to evaluate all the cascade delete logic and any other operations and ensure they are contained to the workspace they are in.

techfg:
# What does this PR do?

Ensures that workspace is respected when collection & integration type metadata is being deleted after a collection/integration type has been removed.
 
Resolves #4801 

# Testing

tested manually and confirmed. 

TBD: Write tests.

### DIFF
--- a/apps/platform-integration-tests/hurl_specs/metadata_list_workspace_context.hurl
+++ b/apps/platform-integration-tests/hurl_specs/metadata_list_workspace_context.hurl
@@ -29,7 +29,7 @@ HTTP 200
 # TODO: Unclear why this assert exists as its fragile and likely does not help in any way since underlying
 # data can change across the test suite.  Evaluate this tests purpose and adjust the assert to be
 # explicit/more meaningful or remove.
-jsonpath "$[*]" count == 29
+jsonpath "$[*]" count == 30
 # Spot check a uesio/tests collection
 jsonpath "$['uesio/tests.animal'].key" == "uesio/tests.animal"
 jsonpath "$['uesio/tests.animal'].namespace" == "uesio/tests"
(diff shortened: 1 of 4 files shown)


apps/platform-integration-tests/hurl_specs/metadata_list_workspace_context.hurl
```diff
@@ -29,7 +29,7 @@ HTTP 200
 # TODO: Unclear why this assert exists as its fragile and likely does not help in any way since underlying
 # data can change across the test suite.  Evaluate this tests purpose and adjust the assert to be
 # explicit/more meaningful or remove.
-jsonpath "$[*]" count == 29
+jsonpath "$[*]" count == 30
```
humandad, line 32: Wait... Were these asserts actually relying on the bug you just fixed to delete a collection? Not that it's fixed, the collection isn't deleted?

techfg: These are asserting against dev workspace.  I changed the test [here](https://github.com/ues-io/uesio/pull/4830/files#diff-8c49097599d8ce906ab7705dde63ed8bc321ddf3951eae8a387a4e5f69f6c681R732) to delete the "duplicate" workspace I created instead of deleting from dev workspace so these counts went up.

humandad (approved):

PR #4833, opened 2025-04-17
[WIP]: chore: improve cascade delete context handling

techfg:
# What does this PR do?

1. Limits elevating session permissions during cascade delete operations to only what is needed by using AnonSiteSession instead of SiteAdminSession unless already SiteAdminSession
2. Refactors requesting workspace permissions to fix caching issues, fail fast and avoid unnecessary processing, always ensure that params has workspaceid, appname and workspacename.

NOTE - This is still a WIP, need to resolve ensuring workspaceid/appname/workspace name in one scenario and also determine if there is a reliable way to clear the caches when workspace info changes in order to cache uniquekey to improve overall perf.

# Testing

ci & e2e will cover.

### DIFF
--- a/apps/platform/pkg/datasource/cascade.go
+++ b/apps/platform/pkg/datasource/cascade.go
@@ -172,9 +172,18 @@ func performCascadeDeletes(op *wire.SaveOp, connection wire.Connection, session 
 
 	// Perform cascade deletes as an Admin to make sure we find all the records we need to
 	// and to avoid unnecessary security logic
-	adminSession := GetSiteAdminSession(session)
+	// Elevate permissions for cascade deletes to make sure we find all the records we need to
+	// and to avoid unnecessary security logic. If we already have a SiteAdmin or Workspace session,
+	// maintain it, else elevate to AnonSession which will grant admin permission set but ensure
+	// we do not have full SiteAdmin privledges (e.g., in some cases, will skip adding WHERE clause
+	// conditions) since we only need rights to all the metadata objects but should maintain
+	// other criteria as appropriate
+	cascadeSession := session
+	if cascadeSession.GetSiteAdminSession() == nil && cascadeSession.GetWorkspaceSession() == nil {
+		cascadeSession = sess.GetAnonSessionFrom(session)
+	}
 
-	deletes, err := getCascadeDeletes(op, connection, adminSession)
+	deletes, err := getCascadeDeletes(op, connection, cascadeSession)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


techfg:
Marking this back to a draft as it creates some issues:

1. When [limiting the elevation of session](https://github.com/ues-io/uesio/pull/4833/files#diff-d52fbe6d0fc77cad51dae9734fb6ac60d5f5984406656abfcea25a65cc8a696dR181), it causes issues downstream in situations where a `uesio/studio.workspae` is automatically added as a condition on loads when not a [SiteAdmin](https://github.com/ues-io/uesio/blob/b9822c2741beb5a1ed413e507c8c84ab7f1f51e7/apps/platform/pkg/bot/systemdialect/systembot_load_allmetadata.go#L143).  In this case, since we do not guarantee a `workspaceid` will always be present, the loads return no records because it applies the condition `uesio/studio.workspace = ""`.  The quick solution here is to revert back to always being a SiteAdmin in cascade deletes but that isn't ideal since we only need that level of perms due to other issues.
2. One solution is to always make the database call [here](https://github.com/ues-io/uesio/pull/4833) when `!haveAccess || wsKeyInfo.HasMissingField()) but this means we are always going to the database every time when we only have uniquekey to start with, efven when `ModifyAllRecords` is true.  Most of the time we have the workspaceid already but in some situations (e.g. viewing a collection detail page), we only have the unique key.  The extra database hits wouldn't be frequent but they would increase.
3. To solve that, we could cache uniquekey but that requires coordination and cache invalidation whenever a workspace is deleted since the uniquekey->workspaceid map is mutable.  See https://github.com/ues-io/uesio/pull/4833/files#diff-317796f4eebb933e1ffd910bda46a65102768bb7695ce969e17d354618014554R38
4. Lastly, the cascades themselves could span one or more workspaces.  For example, bulk deleting from collections (or any other metadata type) from within Studio prod site itself will result in the operations spanning more than a single workspace (assuming items deleted do).  The workspace information in params in this scenario is actually studio and the collection is workspace which is not the collection(s) we are targeting for delete.  However, when deleting an app that has multiple workspaces from within studio itself, the cascade occurs per workspace so in that scenario, the params could be trusted if we had the full complement of them.

The options are:
1. go back to always being a siteadmin in cascadedelete for now until the entire thing can be revisited.  The rest of the changes can remain in place and do introduce some improvements
2. Change to always go to the database when `!hasAccess || wsKeyInfo.HasMissingField` - This should likely solve the issue but it's a DB call every time we only have uniqueid and someone has `ModifyAllRecords` perm (it's a db call regardless if `ModifyAllRecords=false` which is typically the case when we only have the unique key FWIW.
3. Implement caching for uniquekey

The broader question is fully understanding the "params" values, especially when in a "bulk" context.  Specifically, the params are `studio/workspace` when bulk deleting in studio site itself vs. `appname/collectionname` when in studio itself (and this never seems to span multiple workspaces in a single invocation).

PR #4835, opened 2025-04-18
fix: ensure agents are deleted when workspace gets deleted

techfg:
# What does this PR do?

Ensures that agents & their userfiles are physically deleted from db & filesystem when workspace is deleted.

# Testing

Manually tested and confirmed expected behavior.


humandad (approved):
I always forget this!

PR #4836, opened 2025-04-18
feat: add additional sql operators

techfg:
# What does this PR do?

1. Adds support for the following sql operators:
    1. sw = STARTS_WITH (already supported in conditionsapi, only added to deleterecordapi)
    2. cs = CONTAINS  (already supported in conditionsapi, only added to deleterecordapi)
 2. Modified our e2e & integration tests to always create apps starting with the name "tests"
 3. Modified our test cleanup code to delete all apps starting with "tests" name
 4. Fixed a few issues with asserts on existing tests that were not correctly querying for data and/or retrieving files so they would never return any data (even if it was there) but were asserting on counts = 0.  If there is data, it will be retrieved now so the asserts will work as expected.

> [!IMPORTANT]
> There is some down side to adding support for these operators.  These operators are now exposed to the deleterecordapi meaning you could delete a lot of stuff all at once with a simple http DELETE.  This is helpful, as demonstrated by how it helps us cleanup tests, but does cause some pause.  There are other options for fully cleaning up our tests if we do not want to add these operators due to performance reasons.
 
# Testing

Tested all operators manually and all work as expected.  ci & e2e will cover rest.

### DIFF
--- a/apps/platform-e2e/cypress/support/testdata.ts
+++ b/apps/platform-e2e/cypress/support/testdata.ts
@@ -12,7 +12,7 @@ export function getUniqueAppName() {
     { length: 8 },
     getRandomLowercaseAlphabetLetter,
   ).join("")
-  return `e2e${randomLetters}`
+  return `tests_e2e_${randomLetters}`
 }
 
 export function deleteApp(appName: string) {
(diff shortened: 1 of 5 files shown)


humandad (approved):

humandad (changes requested):
My only concern is that the naming of the "LIKE" and "ILIKE" conditions seem very "postgres specific" and have a "postgres specific" syntax (using the `%` character as the wildcard). Down the road, implementing these condition operators for other adapters could be problematic.

It doesn't seem like you even need these operators to accomplish what you want here anyways. Could we just drop LIKE and ILIKE and keep the rest of this PR?

I also don't see LIKE and ILIKE used in the tests or

humandad (approved):

techfg:
Agreed.  Removed LIKE/ILIKE, we can revisit in future if/when requirement arises.

PR #4838, opened 2025-04-18
add chat panel feature flag

humandad:
# What does this PR do?

1. Removes unnecessary loading of all feature flags and language labels in the builder dependencies. (This isn't necessary, because this is done anyways for *every* page load whether or not you're in the builder. Doing this in the builder, was essentially loading these things twice.

2. Adds the `uesio/studio.chat_panel` feature flag.

3. Makes sure that the `uesio/studio.chat_panel` feature flag value is available in build mode.

4. Removes all `selectByName` selectors. (for feature flag and select list). This is a naive approach to getting values. You can't just ask for a value by name without the fully qualified key. How would you differentiate values from different namespaces?

5. To request a feature flag value, you now need to provide the fully qualified key. If you don't, it will use the view in context's namespace as the namespace of the key you're requesting.

### DIFF
--- a/apps/platform/pkg/routing/deps.go
+++ b/apps/platform/pkg/routing/deps.go
@@ -501,8 +501,6 @@ func GetBuilderDependencies(viewNamespace, viewName string, deps *preload.Preloa
 		deps.ComponentVariant.AddItem(variants[i])
 	}
 
-	// Load in studio specific feature flags.
-
 	deps.Component.AddItem(preload.NewComponentMergeData(GetBuildModeKey(builderComponentID), true))
 	deps.Component.AddItem(preload.NewComponentMergeData(GetIndexPanelKey(builderComponentID), true))
 
--- a/apps/platform/pkg/routing/deps.go
+++ b/apps/platform/pkg/routing/deps.go
@@ -416,12 +416,21 @@ func GetWorkspaceModeDeps(deps *preload.PreloadMetadata, session *sess.Session, 
 		return err
 	}
 
-	err = bundle.Load(theme, nil, session.RemoveWorkspaceContext(), nil)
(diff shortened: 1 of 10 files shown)


apps/platform/pkg/routing/deps.go
```diff
@@ -416,13 +416,22 @@ func GetWorkspaceModeDeps(deps *preload.PreloadMetadata, session *sess.Session,
...
+
+	err = bundle.Load(theme, nil, baseStudioSession, nil)
 	if err != nil {
 		return err
 	}
 
 	deps.Theme.AddItem(theme)
 
+	// Add in any builder-specfic feature flags.
+	chatPanelFlag, err := featureflagstore.GetFeatureFlag("uesio/studio.chat_panel", baseStudioSession, baseStudioSession.GetContextUser().ID)
```
techfg, line 429: Is there a way to identify all the builder specific feature flags dynamically so that we don't have to add a line here ever time we add one?  Even if it's just a naming convention maybe?

techfg, line 429: Previous code was getting an Anon session here, do we still need that?

humandad: I thought about just adding them all, but I figured it would be better to just add the ones we need. I guess we could have a naming convention. We could also move the feature flag to the uesio/builder namespace. (but I actually started with that and changed that once to the uesio/studio namespace)

techfg: yeah, a namespace or naming convention that we can use a Starts_with or like type of lookup would work.  I'm good with leaving this as-is and revisiting if/when we end up with more that are builder specific.

techfg (approved):

Issue #4951, opened 2025-06-27
bug: detection of client/remote ip has several shortcomings


techfg:
The [remote ip](https://github.com/ues-io/uesio/blob/main/apps/platform/pkg/middleware/request_context_injector.go#L62) is currently derived and included in the logs, however it has several shortcomings.  For example:

1. It uses the left-most ip value which cannot be trusted and is easily spoofed
2. It does not factor in private addresses
3. Determining which ip to "choose" is not configurable

Depending on the network configuration that the platform is running within, determining the client/remote ip can vary (e.g., how many trusted proxies are between the "client" and the "server".  For purposes of logging/rate-limiting/etc. and in the interest of security, the client ip should come from a "trusted" source.

The following should be implemented:

1. Ensure client ip comes from a trusted source
2. Add configuration options to the platform to determine which "strategy" to use for determining client/remote ip

Reference: 
- https://adam-p.ca/blog/2022/03/x-forwarded-for/
- https://github.com/realclientip/realclientip-go
- https://github.com/go-chi/chi/pull/967

PR #4952, opened 2025-06-27
chore: improve request logging

techfg:
# What does this PR do?

Improves request logging and middleware to remediate some limitations/issues with the previous approach:

1. Request logging was not applied to all routes (e.g., static routes)
2. Authentication was not applied to all routes (e.g., static routes)
3. Requests would not be logged if a failure occurred prior to the logging middleware triggering
4. Lots of custom code around log output format, especially in dev mode
5. Session & Request references were "injected" into context and every log message that had one or both of those in context would be enriched with session/request values - not just the request log itself
6. Although request & session were injected, only two log messages would actually use the "current context" so the values added in step 5 were not used in most cases because a default `context.Background()` was applied
7. No support for distributed/correlated logs (e.g., log messages farther down in the pipeline could not be associated to the actual request that triggered the request)
8. Custom log handler did not properly implement slog.Handler which could result in a different handler being created/passed back and custom logging not occurring
9. Remote IP detection is not accurate - see #4951 
10. PII (e.g., username) was being logged

This PR address most of the above issues:

1. Adds a "traceid" middleware that injects a unique traceid in to context that will be emitted in every log message
2. Eliminates several middlewares with a single `RequestLogger` middleware that is responsible for
    1. Request Logging - Every request is logged assuming UESIO_LOG_LEVEL is set to INFO or lower
    2. Usage Tracking - Adds usage metrics for EGRESS_BYTES, INGRESS_BYTES & REQUEST_COUNT
3. Refactors middleware configuration so that all routes have traceid/requestlogging/authentication
4. PII is not logged
5. Request logs contain information regarding app/site/workspace including name & ids along with userid as well as hostname and other relevant information.

A few notes:
1. The remote ip issue is not addressed but created #4951 to track this
2. A custom version of [httplog](https://github.com/techfg/httplog/tree/custom-release) is currently used via `replace` directive in `go.mod`.  Once all the [issues](https://github.com/go-chi/httplog/issues?q=is%3Aissue%20author%3Atechfg) logged against it are resolved and a new version released, the custom version can be removed in favor of the new `httplog` release.
3. This PR sets the stage for proper distributed tracing and log correlation by introducing a `traceid` but full distributed logging/tracing support needs to be implemented including a configurable back-end provider

# Testing

Tested locally and logs emit as expected.


apps/platform/pkg/cmd/serve.go
```diff
@@ -65,19 +72,36 @@ var versionedItemParam = fmt.Sprintf("%s/%s/%s", nsParam, versionParam, namePara
...
+	// We hang /health of baseRouter because of the way Authentication middleware currently works. If it does not
+	// find a "site" based on the request.Host an HTTP 500 is currently thrown. For proper health checking, we need
+	// to make sure that we can access the root domain since we may not have dns to an actual site in some cases (e.g.,
+	// inside docker container, we can't resolve studio.<domain>.<tld>) without a hosts entry. If/When authentication
+	// is improved to allow for more graceful "no site" scenario, "baseRouter" and "r" could be combined.
+	baseRouter.HandleFunc("/health", controller.Health).Methods(http.MethodGet)
+
+	r := baseRouter.NewRoute().Subrouter()
+	r.Use(
+		// all routes that follow should honor authentication which pay just be a public session
```
humandad, line 95: misspelled comment "pay" -> "may"

humandad (approved):
just one small comment spelling nitpick.

humandad (approved):

techfg:
> My only other comment is that maybe logging session ids could be a bad idea? You can literally log in with those if you set them as your cookie. (Maybe we need a "public" session id and a "private" one. Just wondering what your thoughts are on this.

Yep, we should remove if this is the auth token.  Updated PR, thanks! :)

PR #4981, opened 2025-07-02
Fix slot context issues

humandad:
# What does this PR do?

This PR reverts some of the changes I made in February. https://github.com/ues-io/uesio/pull/4599

1. Pass the context of the creator of the slot as part of the slot definition so that it can be used when it's time to render.
2. For example, if the contents of a slot reference a wire in the same view, it should be available when that slot is rendered. Even if that slot is being rendered as part of a different view.

EXAMPLE:
```
- uesio/core.view:
    view: uesio/cms.admin_nav
    params:
      selected: articles
    slots:
      crumbs:
        - uesio/io.group:
            uesio.variant: uesio/appkit.breadcrumbs
            components:
              - uesio/appkit.icontile:
                  tileVariant: uesio/appkit.breadcrumb
                  title: Articles
                  icon: article
                  signals:
                    - signal: route/NAVIGATE_TO_ASSIGNMENT
                      collection: uesio/cms.article
                      viewtype: list
              - uesio/appkit.icontile:
                  tileVariant: uesio/appkit.breadcrumb
                  avatarVariant: uesio/appkit.breadcrumb
                  title: ${article:uesio/cms.name}
```
In this example, on the last line there is a merge of `${article:uesio/cms.name}` but it happens inside of a view slot.
This will not merge correctly without the changes made in this PR.

However, to keep the fix that https://github.com/ues-io/uesio/pull/4599 fixes, we need to pass on the customSlotLoader from the current rendering context. (This PR also does that)

This is pretty complicated, so I'm happy to discuss on a call.

### DIFF
--- a/libs/ui/src/component/component.spec.tsx
+++ b/libs/ui/src/component/component.spec.tsx
@@ -87,12 +87,14 @@ const componentTypeWithSlotsAndContext = {
   ],
 }
 
-const getViewContext = () =>
-  new Context().addViewFrame({
-    params: { foo: "oof", bar: "rab" },
-    view: viewName,
-    viewDef,
-  })
+const getViewFrame = () => ({
+  params: { foo: "oof", bar: "rab" },
+  view: viewName,
+  viewDef,
+  type: "VIEW",
+})
+
+const getViewContext = () => new Context().addViewFrame(getViewFrame())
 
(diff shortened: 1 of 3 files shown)


techfg (approved):
Recent commit makes it much easier to follow/understand, looks great!

PR #4982, opened 2025-07-02
Update to latest models

humandad:
# What does this PR do?

1. Adds a default model to be used for bedrock when one isn't provided.
2. Uses the default model whenever possible in our calling code.
3. Removes special region handling for Claude 3.5 and switches to cross-region inference models.

Now, most of our AI features will use Claude 3.5 Haiku (Upgraded from Claude 3 Haiku). The builder agent will now use Claude 4 (Upgraded from Claude 3.5 v2).

# Testing (I did each of these things)

1. Test creating fields for a collection
2. Test image generation models in sitekit.
3. Test the sitekit starter template.
4. Test the appkit starter template.
5. Test the view builder agent.

NOTE: I have not requested model access to these new models in the PROD AWS account, only DEV.

### DIFF
--- a/apps/platform/pkg/bot/systemdialect/systembot_listener_runagent.go
+++ b/apps/platform/pkg/bot/systemdialect/systembot_listener_runagent.go
@@ -31,7 +31,7 @@ func runAgentListenerBot(params map[string]any, connection wire.Connection, sess
 		return nil, err
 	}
 
-	modelID := param.GetOptionalString(params, "model", bedrock.CLAUDE_3_5_SONNET_MODEL_ID)
+	modelID := param.GetOptionalString(params, "model", bedrock.CLAUDE_4_SONNET_MODEL_ID)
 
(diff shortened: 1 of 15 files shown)


techfg (approved):
Looks good, really like removing the special case region handling :)

Only comment, and it's a minor nit, is in places where we used to specify a model name (e.g. [here](https://github.com/ues-io/uesio/pull/4982/files#diff-f7c4968c44b43858b25ae7b339ef5844e8677c4f24ebfa34e337b980380e2132L47)) but now rely on the `""` empty model and setting default model handling, my thinking is that it may be best to always be explicit in options rather than rely on default behavior underneath?  In short, just specifying the modelId in the options and using the `BEDROCK_DEFAULT_MODEL_ID` constant.

humandad:
Unfortunately, you can't access constants in go inside of bot code without adding additional functions there. So it would be hard to access the value of the model without just hardcoding it.

PR #4987, opened 2025-07-04
feat: adds the userfile DELETE signal

humandad:
# What does this PR do?

1. Adds a signal "BAND" for working with userfiles.
2. For now we're just adding the "DELETE" signal.


techfg (commented):
The client types haven't been maintained and need a lot of work to match what's actually available/supported but we should add anything new to them I think?

https://github.com/ues-io/uesio/blob/main/libs/ui/src/public_types/client/index.d.ts

humandad:
I don't think specific signal definitions are in there for any signals, so I think for now we're good not updating that file. We definitely need to look at that as part of future work though.

PR #4989, opened 2025-07-04
feat: file attachments view

humandad:
# What does this PR do?

Screenshot 2025-07-04 at 12 58 41 PM

1. Abstracts file attachments for a record into a single subview component.
2. Uses this new view for the Files UI in Workspaces.

# Note

For the functionality of this PR to work, the following PRs must be merged.
https://github.com/ues-io/uesio/pull/4988
https://github.com/ues-io/uesio/pull/4987
https://github.com/ues-io/uesio/pull/4986

# Testing

Try out the Files UI and verify that everything works as expected for both files from the local namespace, as well as files that are installed from a bundle.

### DIFF
--- a/libs/apps/uesio/appkit/bundle/components/tile_file.yaml
+++ b/libs/apps/uesio/appkit/bundle/components/tile_file.yaml
@@ -1,0 +1,93 @@
+name: tile_file
+category: LAYOUT
+type: DECLARATIVE
+definition:
+  - uesio/io.card:
+      uesio.variant: uesio/appkit.main
+      uesio.styleTokens:
+        panelHeader:
+          - p-4
+        titlebarTitle:
+          - break-all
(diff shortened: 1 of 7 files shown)


techfg (approved):
Significant improvement over prior UI/UX, awesome stuff!

The only thing that this does not contemplate and will cause an issue for that I found is mobile since there is no "hover" on mobile.  In reviewing the code, it looks like you are re-using the existing `uesio/appkit.itemaction` and that is where the problem lies so up to you if you want to address it in this PR or create a separate issue to track it and get it resolved.  We have a number of "mobile" issues, this is just one of them.

humandad:
yeah, i'll see if I can fix the hover stuff real quick.

techfg:
FYI - The hover work was moved to #4998 and #4999

PR #4996, opened 2025-07-05
chore: support http caching for favicon

techfg:
# What does this PR do?

Enables http caching for favicon.

Note - This only addresses the ability to support http caching for favicon.  See #4674 (which has been updated to account for the changes in this PR) for other issues related to favicon.

@humandad - When reviewing, please pay particular attention to https://github.com/ues-io/uesio/pull/4996/files#diff-dee56ef84e65aee0654893201db7dae31624d048d65c98e014031ee14fa1d7b9R15.  Previously, the code would `GetContextAppBundle` and `Favicon` from there, however this PR changes to always just use the current site (via `session.GetSite()`).  I believe the net result should be the same but would like to verify.

# Testing

Tested manually and proper favicon url with `v=###` syntax is generated and cache headers set as expected.

### DIFF
--- a/apps/platform/pkg/controller/mergedata.go
+++ b/apps/platform/pkg/controller/mergedata.go
@@ -279,8 +279,8 @@ func GetSiteMergeData(site *meta.Site) *preload.SiteMergeData {
 func getFaviconVersion(site *meta.Site) string {
 	// intentionally ignore error here since it should not happen
 	// and if so we use version of bundle as a fallback
-	fileItem, _ := meta.NewFile(site.GetFavicon())
-	if bundlestore.IsSystemBundle(fileItem.Namespace) {
+	fileItem, err := meta.NewFile(site.GetFavicon())
+	if err == nil && bundlestore.IsSystemBundle(fileItem.Namespace) {
 		return os.Getenv("UESIO_BUILD_VERSION")
 	} else {
 		return site.Bundle.GetVersionString()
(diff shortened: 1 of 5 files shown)


humandad:
On the `GetContextAppBundle` thing, I think your new approach is better and more correct.

apps/platform/pkg/controller/mergedata.go
```diff
@@ -257,15 +257,33 @@ func GetRoutingMergeData(route *meta.Route, metadata *preload.PreloadMetadata, s
...
+// which could have a version different from the site bundle version. However, looking up the favicon
+// to that level has a perf cost (even though it would be cached) that seems unnecessary in the majority
+// of cases. As written below, we will align the favicon caching with the site bundle version which
+// should be sufficient for most use cases. If needed, a specific lookup to the referenced favicon could
+// be performed and the version of the actual favicon source used instead.
+func getFaviconVersion(site *meta.Site) string {
+	// intentionally ignore error here since it should not happen
+	// and if so we use version of bundle as a fallback
+	fileItem, _ := meta.NewFile(site.GetFavicon())
+	if bundlestore.IsSystemBundle(fileItem.Namespace) {
```
humandad, line 283: I think this will panic if there's an error. Maybe we should check to see if fileItem is nil?

techfg: Great catch!  Updated.

humandad (approved):
LGTM

PR #5003, opened 2025-07-09
[WIP]: fix/autonumber-field-must-be-unique

techfg:
# What does this PR do?

Fixes an issue where "AUTONUMBER" fields would not receive unique integer portion of field value when concurrent requests were "inserting" in to db (either on same host or different hosts).  The issue was due to retrieving the current `MAX(autonumber)` from DB, however if two requests were simultaneously executing, they would both receive the same value because neither had saved yet.

In evaluating options for resolution, a few approaches were considered:

1. Redis w/ SET NX & INCR - This would require seeding the initial value retrieved from the database via `SET NX` and then using INCR each time a new value is required.  While fairly straightforward to implement, it does present limitations in a multi-master Redis scenario (we have existing gaps in other areas of the code base with this issue).  The multi-master scenario would require [redlock](https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/) which is the recommended approach for this situation, even single master, currently anyway but that introduces further complexity for little gain.
2. [Twitter Snowflake IDs](https://en.wikipedia.org/wiki/Snowflake_ID) - There are multiple golang packages that support this format which would simplify.  The benefit of it is its speed and that it would maintain an "integer" value similar to the previous `AUTONUMBER` approach.  There would be gaps in the numbers, just like the current approach.  The main downside to this approach is that the "machine ID" must be unique and managing this across all environments that would "host" uesio may prove challenging (but solveable).
3. UUID v7 - Relatively straightforward approach but two downsides: 1) 32 characters; 2) No way to differentiate between a "record id" (for our actual record id) and an auto-number id.  While there shouldn't be a need for this, having them be different format does provide some level of clarity.  THe benefit is that no configuration per "farm" or "instance" is required.
4. UUID v7 Base 58 Encoded - All the benefits of option 3 but only 21 or 22 characters

The idea behind "AUTONUMBER" was to provide a method for "External IDs" to be easily generated when needed, but more importantly, be able to support import/export.  While there is some value in having sequential integers starting at 1, once a collection grows in size, this benefit fades and doesn't provide a lot of value to begin with.

Given the goals of "AUTONUMBER", Option 4 (UUID v7 Base 58 Encoded) has been implemented which will result in `CQvps39NuxG1y2SUgZEKW`

# Testing

Tested locally with multiple autonumber fields and all works as expected.  ci & e2e will cover.

TBD:
1. Existing `autonumber` field in DB - This field was only used to be able to obtain the "next" id and is no longer used as of this PR.  However, dropping it (and it's unique index) would mean it would be incredibly difficult to "go back" to previous database state because the record data itself would need to be parsed based on the metadata for any autonumber fields in the collection and then the "id value" extracted.  This is further complicated by the fact that the prefix can change over time across bundle versions.  We could just drop the field & index and never go back (which would be unlikely to need anyway) or convert the field to a string for now and continue to populate and remove both at a later time.  DB Migration scripts need to be adjusted accordingly.
2. Should we consider a "Format" option instead of "Prefix" allowing for any format for things like suffixes, etc. and also avoiding the "fixed" `-` that we currently have.  The "Format" field can be blank (which would just result in the auto id being set) or any text format that contains `{id}` token where the "auto id" would go.
3. To deploy, db migrations must be run immediately after platform is released - this needs to be coordinated with an separate one-off AWS task per README.  Separate, but related, is to create an issue to improve the db migrations to automatically run upon start-up if needed - this needs to be coordinated itself in a multi-host scenario via a distributed lock of sorts (e.g., Redis redlock possibly).
4. "Id/ID" labels - Currently inconsistent with the use of "Id" and "ID" for labels, method names, variable names, etc. - need to identify a standard and begin to adjust for consistency throughout entire codebase.


techfg:
Updates per TBD items:

1. Dropping existing `autonumber` field & its unique index in DB - there is no easy way to go back but its unlikely that this would ever be needed
2. Added "Format" option and deprecated "Prefix" - Any `{id}` tokens in the format will be replaced with the autoid value
3. DB migration for dev/prod - Will be done as a one-off until a better "migration" automation can be implemented
4. Standardizing on `Id` for UI label text cases - need to define a style guide

A couple of other notes:

1. Custom marshallers (JSON & YAML) are used in AutonumberFieldMetadata to handle the legacy `Prefix` property
2. Each `AUTONUMBER` field in a collection will receive a unique `auto id` - previously, all `AUTONUMBER` fields were given the same `auto number` value generated for the collection record.

humandad (approved):
LGTM

PR #5030, opened 2025-07-15
refactor: use anthropic-go-sdk where possible

humandad:
# What does this PR do?

This PR significantly refactors and simplifies the `bedrock` package in our backend code.

1. Use the official anthropic go sdk for message api inputs and outputs
2. Use the official anthropic go sdk to call bedrock for regular inference and streaming.
3. Move usage handling to the individual model handlers
4. Move options parsing to the individual model handlers

# Testing

1. Use the chat panel in the builder to modify a view.
2. Generate new fields on a collection. (they should stream in)
5. Generate sample data for a collection. (those records should stream in as well)
6. Create a new app using the sitekit starter template. (It should work the same as before)
7. Create a new app using the appkit starter template. (It should work the same as before)


apps/platform/pkg/bot/systemdialect/systembot_listener_runagent.go
```diff
@@ -101,7 +106,7 @@ func runAgentListenerBot(params map[string]any, connection wire.Connection, sess
 	resultMessages := []anthropic.ContentBlockUnion{}
 	err = datasource.HydrateOptions(result, &resultMessages)
 	if err != nil {
-		return nil, exceptions.NewBadRequestException("invalid message format for agent", nil)
+		return nil, exceptions.NewBadRequestException("invalid message format for agent: "+err.Error(), nil)
```
techfg, line 109: any reason you didn't just pass err as 2nd param?

humandad: oh, nope. I'll fix that

techfg (approved):

PR #5045, opened 2025-07-21
chore: migrate to scs for session management

techfg:
> [!WARNING]
> __BREAKING CHANGE__: Support for `filesystem` for `UESIO_SESSION_STORE` has been removed. You must migrate to `redis` or the default of `memory`.

# What does this PR do?

Migrates session management from https://github.com/icza/session to https://github.com/alexedwards/scs.

This PR's goal is to migrate to scs with 100% functional equivalence to the approach that existed with `icza`. With a few minor exceptions, the code was left as-is in terms of its structure and placement but the functional result is identical before/after. There are several more changes that should be made (including a fairly significant refactoring of the code) but breaking this up into pieces so that its easy to identify the migration itself vs. the forthcoming improvements.

The rationale behind this migration is a few-fold but it stems from the fact that we were not updating the "Last Accessed" time in our session meaning that the 12-hour expiration that we set was absolute despite it being intended to be sliding. This would be easy enough to solve by updating the accessed time and saving the session back to the store at the end of each request.  However, icza's does not have a "session" or "store" specific API to "update" a single session in the backing store. The approach it offers is to have a "SessionManager" per request and "Close" the session manager at request completion which would then flush the current session(s) being managed by that manager to the store. 

While making this change would have been possible, there would have been a decent amount of refactoring of the code to accomplish and in evaluating that, a few other related issues were identified:

1. icza hasn't has a new release since August 2024 and its unclear how active its maintenance is
2. icza API is largely built to be "memory" backed and methods like "Add" do not have a signature that returns an `error` making it more difficult to handle errors when saving to the backing store like `redis`
3. icza is lacking a lot of modern features (e.g., SameSite, efficient encoding of data, multiple sessions per request, etc.) that other libraries offer

Given these limitations, two packages were evaluated for migration:

1. https://github.com/gorilla/session
2. https://github.com/alexedwards/scs

Both of these packages offer similar functionality and have robust communities. `scs` was chosen largely because while we use [gorilla/mux](https://github.com/gorilla/mux) for our mux, gorilla appears to no longer be actively maintained, and we have plans to move away from it as a result so does not make sense to use `gorilla/session` for same reason.

Bottom Line: 100% feature equivalence but we should see an overall boost in performance due to using [gob](https://github.com/alexedwards/scs?tab=readme-ov-file#working-with-session-data) for encoding vs. JSON (see https://alecthomas.github.io/go_serialization_benchmarks/).

# Testing

ci & e2e will cover.


techfg:
> I don't see any issues with the approach here. Not sure why the tests are failing though.

Thanks for taking a look.  Yeah, I'm not sure why the tests are failing, they don't locally although I have been able to get a similar yet different issue which I think underneath is the same if I have devtools open while running the tests locally. In short, the browser is sending a different session ID which is resulting in a new public session to be created. It's getting the "new" session ID because it's making a call for the javascript sourcemap without using the session id cookie it already has which results in a new one.  No clue yet why this is happening given the changes as I can't reproduce it with `main`.  Reading some stuff about Chromium browsers not passing cookies through on source maps but again, why did it work before.  Continuing to debug.... :(

techfg:
Closing in favor of #5073

PR #5058, opened 2025-07-23
fix: always generate redirectpath on login events server-side

techfg:
# What does this PR do?

Modifies "login responses" to always contain a `redirectPath` eliminating the `routeNamespace/routeName` usage.

The reason for this change is a few-fold:

1. We always hit the server to process a "login related' event but then, unless an `r=<path>` was specified, would send the route namespace/name to the client who would then go back to the server to get the path and then `pushState`.  This 2nd call to get information we could already obtain impacts performance unnecessarily.
2. While we are processing the "login related event", we are changing the session (auth) token so any inflight requests from the browser may fail depending on when they get processed through middleware.  There is actually another bug (PR coming) where we don't "remove" the "old session" but assuming we were doing that as we should, the inflight request would now have an invalid token.  By transitioning to using `window.location=<path>` instead of `pushState`, we help mitigate this risk because the browser should cancel in-pending requests rather than try to complete them.  Note that there are still situations where an in-process request could have a token that is now expired (this seems to occur most commonly with source map requests when dev tools is open) but forcing a full page refresh minimizes all these risks
3. We want to ensure that the full page, all resources, etc. are loaded in the context of the "now current" user - a full page refresh is the only way to achieve this

NOTE: There are more PRs coming in/around the security of session management, etc. as mentioned above.  This is not final state but the next step.

# Testing

Tested manually and confirmed expected behavior.  e2e & ci will cover rest.

### DIFF
--- a/apps/platform/pkg/auth/login.go
+++ b/apps/platform/pkg/auth/login.go
@@ -1,7 +1,6 @@
 package auth
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -12,6 +11,7 @@ import (
 	"github.com/thecloudmasters/uesio/pkg/controller/filejson"
 	"github.com/thecloudmasters/uesio/pkg/datasource"
 	"github.com/thecloudmasters/uesio/pkg/preload"
+	"github.com/thecloudmasters/uesio/pkg/routing"
 	"github.com/thecloudmasters/uesio/pkg/types/exceptions"
 	"github.com/thecloudmasters/uesio/pkg/types/wire"
 
@@ -28,7 +28,7 @@ func GetResetPasswordRedirectResponse(w http.ResponseWriter, r *http.Request, us
 
 	redirectPath := redirect + "?code=" + code + "&username=" + username
 
-	return NewLoginResponse(preload.GetUserMergeData(session), session.GetSessionID(), redirectPath, "", ""), nil
+	return NewLoginResponse(preload.GetUserMergeData(session), session.GetSessionID(), redirectPath), nil
 }
 
 func GetLoginRedirectResponse(w http.ResponseWriter, r *http.Request, user *meta.User, session *sess.Session) (*LoginResponse, error) {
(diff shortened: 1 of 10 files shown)


apps/platform/pkg/auth/login.go
```diff
@@ -59,20 +52,18 @@ func GetLoginRedirectResponse(w http.ResponseWriter, r *http.Request, user *meta
...
-
 	if redirectPath == "" {
-		if redirectKey == "" {
-			return nil, errors.New("no redirect route specified")
+		route, err := routing.GetUserHomeRoute(user, session)
+		if err != nil {
+			return nil, err
 		}
-		redirectNamespace, redirectRoute, err = meta.ParseKey(redirectKey)
+		redirectPath, err = url.JoinPath("/", route.Path)
```
humandad, line 60: I assume this is fixed in a later PR as we discussed.

apps/platform/pkg/controller/signup.go
```diff
@@ -52,13 +53,18 @@ func Signup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	redirectRouteNamespace, redirectRouteName, err := meta.ParseKey(signupMethod.LandingRoute)
+	route, err := routing.GetRouteFromKey(signupMethod.LandingRoute, systemSession)
+	if err != nil {
+		ctlutil.HandleError(r.Context(), w, err)
+		return
+	}
+	redirectPath, err := url.JoinPath("/", route.Path)
```
humandad, line 61: Same here. I assume this is fixed with the new getRouteUrlPrefix function

humandad:
✅

apps/platform/pkg/auth/login.go
techfg: Yep, #5070

apps/platform/pkg/controller/signup.go
techfg: Yep, #5070

PR #5059, opened 2025-07-23
fix: signup should redirect with public session

techfg:
# What does this PR do?

Changes the behavior of post-processing of signup to eliminate the user of system session when evaluating & processing redirect.  

We should never be redirecting using a system session for a number of reasons, the least of which is that we send user information in the redirect along with the session token.  Fortunately, these values were explicitly "empty string" but the use of system session for anything but explicit system operations should be avoided as its prone to (unexpected) mis-use in downstream processing.

When a user signs up, the session should be treated as public.

NOTE: There is still more work to be done on improving these flows and control current "logged in/out" behavior.  One thing in particular is that we currently allow login & signup to occur even when the current user is logged in.  Since we only support a single user session at a time currently, accessing login or signup should redirect to "home" or "site root" and not allow them to continue.  However, due to the way routing works for site/local routes, there is no straightforward way to detect if the route is a signup/login route (even the site settings aren't enough because any route could be signup/login beyond the default signup/login). In forthcoming PRs, every "login type" of activity (login/logout/signup/etc.) will force a logout of any current user to eliminate the duplicate users and ensure that any auth tokens are removed immediately (they are currently orphaned in these flows).

# Testing

Tested manually and confirmed expected behavior. ci & e2e will cover basics.

### DIFF
--- a/apps/platform/pkg/controller/signup.go
+++ b/apps/platform/pkg/controller/signup.go
@@ -48,12 +48,44 @@ func Signup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// we currently support signup even if you are a logged in user so we can't use the session we received
+	// since we are acting on behalf of the "new user" and not the "current session user" at this point.
+	// We should not allow signup (or login) for users that are already logged in but there is straightforward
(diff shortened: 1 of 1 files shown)


humandad:
✅

PR #5061, opened 2025-07-24
chore: improve timing of nitial health detection of worker

techfg:
# What does this PR do?

Healthcheck job won't run until 30 secs after start so write the file as soon as we start to improve initial health detection by docker/ecs/etc.

# Testing

Tested manually and confirmed.


humandad:
✅

PR #5063, opened 2025-07-25
fix: ensure browser session is removed on priviledge changes

techfg:
# What does this PR do?

1. Ensures that when a privilege change occurs (e.g., login, signup, etc.) that the previous browser session is removed. Previously, a new browser session with corresponding token was created, however the previous browser session was orphaned meaning it would stay and be available in cache for the remainder of its absolute lifetime of 12 hours
2. Ensures that whenever a "non-system specific" privilege change is occurring (e.g,, login, signup), it is performed as the Public User. Previously, the action was performed in the "current user" context, however since we allow "logged in users" to access login, signup, etc. (we can't reliably redirect away because any route could be a login/signup/etc. route), we were processing them as the current user. While this was largely protected from "leaking" and/or "using" the current user information in anyway, it's prone to (unintentional) mis-use and needs to be avoided.

As mentioned in recent PRs, there are several more coming related to improving and ultimately refactoring the auth flows.  This is just the next step of the process and should not be considered permanent state.

# Testing

Tested all scenarios manually and confirmed. e2e & ci will cover rest.

### DIFF
--- a/apps/platform/pkg/auth/cli_token.go
+++ b/apps/platform/pkg/auth/cli_token.go
@@ -95,7 +95,7 @@ func CLIToken(w http.ResponseWriter, r *http.Request, requestingSession *sess.Se
 		ctlutil.HandleError(r.Context(), w, err)
 		return
 	}
-	browserSession := sess.CreateBrowserSession(w, user, site)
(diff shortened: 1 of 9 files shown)


humandad:
✅

PR #5065, opened 2025-07-25
chore: refactor location of auth related code

techfg:
# What does this PR do?

No functional changes of any kind.  Just refactoring the location of some of the auth related code as the next step (many more yet to come) of improving auth flows.  Focused on putting code in packages and files were it more appropriately fits than its prior location.

Major win here is decoupling sess package from its dependencies on icza and eliminating it from having "controller" functions. Also, eliminating some middleware only functions from auth package.

# Testing

ci & e2e will cover.


humandad:
✅

PR #5070, opened 2025-07-25
chore: handle site & workspace routes in login response redirectPath

techfg:
# What does this PR do?

In #5058, changes were made to derive the `redirectPath` on the server rather than, in some situations, send the route namespace & name back to the client where it, in turn, would call the `/key` API to get the route information and then use it to derive the redirectPath. While it was building the final path in [navigateToRoute](https://github.com/ues-io/uesio/blob/main/libs/ui/src/bands/route/operations.ts#L76), it would determine if a prefix should be applied.

The code in #5058 did not replicate the prefix calculation so this PR closes that gap.

Note that there are some questions about how the client was doing this, if it's doing it correctly and therefore how we should be doing this correctly on the server.  Hence, marking this as [WIP].

# Testing


humandad (approved):
This looks good to me. I think eventually, any redirect to a route by that route's "key" should use this function. The code in Login.go should also use this logic, but that can be a refactor for a later PR.

humandad (approved):

PR #5071, opened 2025-07-26
chore: explicit auth token on uesio session

techfg:
> [!WARNING]
> __BREAKING CHANGE__: The `Session API` function `getId` has been renamed to `getAuthToken`. Any bot related code that utilized `getId` will need to be updated to `getAuthToken`.

# What does this PR do?

Eliminates the "uesio session" `ID` property and adds an `token` property for authentication token.

Previously, the [uesio session](https://github.com/ues-io/uesio/blob/main/apps/platform/pkg/sess/session.go) overloaded the meaning of its `Id` property. Sometimes it was blank, sometimes it was a name (e.g., `wsbundlecache`), sometimes it was a guid and most times, it was an authentication token. What was true all the time is that it was confusing, difficult to discern when to use what and most importantly, put security at risk (e.g., someone could unknowingly log the `ID` thinking it was just an `ID` when really it was an auth token - I did this myself!).

This PR eliminates the `ID` completely primarily because there really is no value in having an `ID` for each session currently. It would be helpful for logging purposes possibly but the current code structure doesn't lend itself to traceability rendering an `ID` pretty much useless in logs. There is the recently added `traceID` in request context now and that should really become the way to trace all requests through their lifecycle (user initiated or background). Work on traceability will be forthcoming but as it stands, session `ID` isn't needed for anything other than logging and traceid should be used instead so `ID` is now gone.

For the situations where `ID` represented an authentication token, a new `GetAuthToken()` function has been added to `Session`. In the cases where `ID` was an auth token, `GetAuthToken()` should be used instead and its intent and meaning is now very clear within the code base.  Along that line, there is something to be said for session not even needing to carry the auth token at all. Based on discussions with @humandad, the main reason it was added was largely to support writing tests so that a test could call an API on the same site avoiding any need of internet connectivity.  That said, there are real world scenarios for allowing bots to call APIs but to my knowledge there are no current use cases for this. For now, leaving the functionality for backwards compat but the inclusion of this should be evaluated and removed if no valid use cases or if there are some, the implementation revisited to ensure it meet the business requirements and does so with security in mind.

# Testing

ci & e2e will cover.

### DIFF
--- a/apps/cli/pkg/auth/check.go
+++ b/apps/cli/pkg/auth/check.go
@@ -9,19 +9,19 @@ import (
 
 func Check() (*preload.UserMergeData, error) {
 
-	sessid, err := config.GetSessionID()
+	token, err := config.GetToken()
 	if err != nil {
 		return nil, err
 	}
 
 	// If there's no current session id stored, no need to make check call
-	if sessid == "" {
+	if token == "" {
 		return nil, nil
 	}
 
 	var userResponse auth.UserResponse
 
-	err = call.GetJSON("site/auth/check", sessid, &userResponse)
+	err = call.GetJSON("site/auth/check", token, &userResponse)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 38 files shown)


humandad:
✅

PR #5072, opened 2025-07-26
fix: cli auth check during login

techfg:
# What does this PR do?

In #5033, a "profile" check was removed from the start of login flow and instead, if we had a current user, we just no-op'd the login. However, this created an issue whereby if there is a `token` in config but that token is no longer valid (e.g., it expired), the auth check API returns a "Guest" user. The change in logic meant that we would always get a user back (barring any unforseen failure) and never proceed with login.  The only way to login was to ensure that the local config did not have a 'token' value.

This PR resolves that issue by eliminating the "check" completely.  It's really unnecessary as the user is asking us to login so we should do as they ask - this is the way the majority of CLI login processes work. The alternative is to ask them to logout first if we have an existing token but that seems cumbersome for the user - just proceed with login and if they successfully log in, replace the token in config with the new one (and logout the old token ignoring any failure in doing so).

In addition to solving this issue, the "Authorization Result" HTML was updated to include instructions for users on how to log in as a different user.  If the user is already logged in via their browser, the current flow will go straight through to successful login and update the local token with a token for CLI so there is no way for the user to "choose" who they want to login as. To solve this for now, text was added to the HTML output to inform the user that they should visit the site, logout and then login again via the CLI. This flow is less than ideal but at least provides the user the information needed. This flow should be improved.

# Testing

Tested manually and confirmed all expected behaviors. ci & e2e will cover basics.

### DIFF
--- a/apps/cli/pkg/auth/login.go
+++ b/apps/cli/pkg/auth/login.go
@@ -174,7 +174,7 @@ var browserHandler = &LoginMethodHandler{
 			}
 		})
 		eg.Go(func() error {
-			code, redirectURL, err := receiveCodeViaLocalServer(ctx, &localServerConfig{readyChan: ready, state: state, codeChallenge: codeChallenge, codeChallengeMethod: "S256", authURL: authURL})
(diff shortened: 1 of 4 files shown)


humandad:
✅

PR #5074, opened 2025-07-26
chore: protect redis namespace registration

techfg:
# What does this PR do?

Ensures registering a redis namespace is synchronized/protected from concurrent access.

# Testing

ci & e2e will cover


humandad:
✅

PR #5075, opened 2025-07-26
chore: samesite should use the default of Lax

techfg:
# What does this PR do?

Cookie SameSite setting should use the default of Lax. Forgot to remove testing code.

# Testing

Manually confirmed Lax is set for the cookie. ci & e2e will cover rest.


humandad:
✅

PR #5076, opened 2025-07-26
chore: do not create browser sessions for public user

techfg:
# What does this PR do?

Eliminates creating a browser session for public users.

Previously, we would create a browser session for public users which included storing the siteid along with the public userid in the session cache. However, there are several downsides to this:

1) Every request must go to the cache to get the public user information
2) Every request must retrieve the user from the user cache

This happens on every request (page, js, css, font, etc.) and has no value because all we are doing is storing the information that we know anyway via `GetPublicUser`.

By eliminating creating a browser session for public user, we avoid all the unnecessary session cache operations which will improve performance of the site for public users as well as reduce utilization/cost of the session store (e.g., redis).

# Testing

Tested manually and confirmed no sessions are created as expected. ci & e2e will cover basics.


humandad:
✅ Nice!

PR #5078, opened 2025-07-27
fix: sychronize creating platform file connection

techfg:
# What does this PR do?

Ensures that creation of platform file connection is properly synchronized/protected from concurrent access.

# Testing

ci & e2e will cover.


humandad:
✅ 
Just curious. Did you ever run into a race condition here? Or did you just notice the possibility of one?

PR #5079, opened 2025-07-27
chore: ensure session always has a context

techfg:
# What does this PR do?

Ensures that session always has a context.

Previously, while a session normally had a non-nil context set via `SetGoContext`, it did not always have a context.  The context property was added to Session in [this commit](https://github.com/ues-io/uesio/pull/3619/files#diff-b50d36d253129bd263753674a60b24374b3b91b19cd0b492aa4c0d700bf1c885R28). As the comment in that commit indicates, it was added as an alternative to changing all the APIs to accept a context.  

The approach taken, while understandable, has led to the code morphing and having a couple of significant shortcomings:

1. Session doesn't always have a context
2. Session passed around during request processing can change (e.g., ensurePublicSession) which could result, mainly unintentionally, in multiple contexts involved in the processing of a request
3. Connection also has an embedded context and often session and connection are passed to the same function - so which one to use?  While they, in theory, should be the same, there's no way to guarantee this so again, multiple contexts can be involved in request processing
4. PlatformBundleStore maintains a singleton connection which has a background context. This causes issues because any code that uses connection.context from it will be using a different context than the session context which came from the request context - again, multiple contexts involved in processing a request

In short, the APIs need to change and explicitly receive a context that should be used so that an entire request flow is using the original context (or background context if a non-HTTP request).

This PR is the first step of several to get to a refactored API by starting with ensuring that every Session has a context which should be the first choice downstream when available.

# Testing

e2e and ci will cover.

### DIFF
--- a/apps/platform/pkg/auth/session.go
+++ b/apps/platform/pkg/auth/session.go
@@ -89,8 +89,8 @@ func GetUserFromAuthToken(token string, site *meta.Site) (*meta.User, error) {
 	return GetUserByID(loginmethod.User.ID, adminSession, nil)
 }
 
-func GetSessionFromUser(user *meta.User, site *meta.Site, token string) (*sess.Session, error) {
-	s := sess.NewWithAuthToken(user, site, token)
+func GetSessionFromUser(ctx context.Context, user *meta.User, site *meta.Site, token string) (*sess.Session, error) {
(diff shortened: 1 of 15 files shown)


humandad:
✅ This is great. It's good to see us handling contexts better!

PR #5080, opened 2025-07-28
chore: refactor login flows

techfg:
# What does this PR do?

Refactors all of the "login" related flows, moving the "controller" portions out of "auth" package and back in to "controller" package.

In #3881, work was done to implement SAML authentication for a POC project. In that PR and some subsequent SAML related PRs, orchestration type logic was moved out of "controller" package and in to "auth." Then, over time, this trend continued resulting in the "auth" package serving multiple purposes - in short, we had "controller/orchestration" type logic in "auth", "controller", "middleware" and "sess".  For example, the "auth" package maintains the "BrowserSession" management logic which only middleware should need access to, however due to the orchestrations and cross-package dependencies, the only place it can live is in "auth" to avoid import cycle errors.

In a prior PR, "sess" was refactored so this PR is focused on the first step of refactoring "auth". Beyond the "single responsibility" theory, the challenge with having orchestration logic in "auth" where its functions accept `http.Request/Response` parameters is that it makes it more difficult to orchestrate different types of slows (e.g., saml, CLI, standrad login, oauth, etc.) as the system grows.  By separating concerns and letting "auth", well "auth", controllers orchestrate and middleware, well "middleware", we achieve a much easier to maintain code base.

Subsequent PRs will target "signup", "password reset", etc., this PR is just focused on all the various login flows.

# Testing

e2e & ci will cover.


humandad:
✅

PR #5082, opened 2025-07-28
chore: improve cli login

techfg:
# What does this PR do?

With the recent refactoring of auth flows, there is an opportunity to improve the CLI login for `mock/platform` logins to target the CLI specifically as the client.  Previously, these flows called the same endpoints as the web based login flows and therefore received responses as if they were a web client - but the CLI is not a web client :)  The refactoring provides the opportunity to orchestrate a different flow for the CLI.  One example of why this was needed is because the web based login returns an HTTP 200 with `redirectPath` when a users account requires a password reset.  The CLI doesn't expect a redirectPath and since the status was 200, didn't know what to do since there was no token returned.

This PR does three things:
1. Adds new a new `auth/cli/<provider>/login` route for CLI logins and returns proper results, including meaningful errors that can be evaluated by the CLI to surface to the user.
2. Removes the `token` property in the standard login response body for the web based login - including it in the body was never necessary other than for the CLI since it shared the exact same login orchestration.
3. Improves the output of error messages in the CLI when making http requests

# Testing

tested manually and confirmed error messages and successful login as expected. ci & e2e will cover rest.

### DIFF
--- a/apps/cli/pkg/auth/login.go
+++ b/apps/cli/pkg/auth/login.go
@@ -234,7 +234,7 @@ func processDirectLogin(method string, payload map[string]string) (*auth.TokenRe
 		return nil, err
 	}
 
-	url := fmt.Sprintf("site/auth/%s/%s/login", methodNamespace, methodName)
+	url := fmt.Sprintf("site/auth/cli/%s/%s/login", methodNamespace, methodName)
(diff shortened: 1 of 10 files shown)


humandad:
✅

PR #5090, opened 2025-07-28
chore: add traceid to background contexts

techfg:
# What does this PR do?

What this does: Adds a traceid for log correlation to places where we use a background context.

What this does NOT do: 
1. Fix the underlying issues with use of background context - in most places that we currently use one, we should not be and instead using session.Context()/request.Context()
2. Ensure that slog statements are using the *Context variants so that the traceid will actually appear in the logs

More PRs coming on improving the flow and use of contexts, this is just the next step.

# Testing

ci & e2e will cover.


humandad:
✅

PR #5094, opened 2025-07-30
chore: add context to log messages

techfg:
# What does this PR do?

1. Adds `Context` to log messages - all `slog` calls now use the `XXXContext` variant with the exception of two (2) calls to `slog.Error` where it would not have been possible to obtain a context to apply without some significant refactoring.  So we're not 100% but we're pretty close now 😄 
2. Refactors platform `cmd` package to create a context with traceid in `main()` and flow that context to commands so that start to finish on a cmd, we have a single traceid - http requests, background jobs, etc. will have their own traceid.

# What this PR does not do?

Address some of the ways in which we "ignore" errors, log a message and return a value instead of a typed error so that higher level functions can evaluate how to handle. We make some assumptions in certain places in the code that shouldn't be made at the lowest of levels. This results in having to pass "Context" much farther down in the call stack then normally necessary.  Future PRs will improve around this and then context can stay at the higher levels and not passed to helper functions.

# Testing

Tested manually and confirmed traceid is logged with every log message.

### DIFF
--- a/apps/platform/main.go
+++ b/apps/platform/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"context"
 	"log/slog"
 	"mime"
 	"os"
 
+	"github.com/go-chi/traceid"
 	"github.com/thecloudmasters/uesio/pkg/bot"
 	"github.com/thecloudmasters/uesio/pkg/usage/usage_memory"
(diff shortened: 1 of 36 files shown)


humandad:
✅

PR #5095, opened 2025-07-30
chore: use current context instead of context.Background

techfg:
# What does this PR do?

Replace use of `context.Background` with the `current context` when its available.  The only things really remaining that use `context.Background` are:

1. Singletons
2. Tests
3. Scheduled jobs

# Testing

ci & e2e will cover.


humandad:
✅ It's so nice to see all this getting cleaned up.

PR #5098, opened 2025-08-01
feat: add file d/l button to files

### LINKED ISSUE #4895: Change: Clicking on a file download replaces the view with the download preview
After a user uploades a file to the image component and the clicks on the file to download the current view is replaced with a preview of the downloaded file forcing the user to either page back or find a way to log in again.

Image

Image

techfg:
# What does this PR do?

In some recent PRs (e.g., #4989), the UI was significantly improved related to displaying files/attachments/file fields. In that process, the "download file" button was lost in certain situations.

1. Adds a "download file" button to `tile_file`
2. Changes the behavior of "downloading a file" on `file fields` and `tile_file` to use content-disposition of attachment

# Testing

Manually confirmed downloads as attachment vs. inline.

Closed #4895


humandad (approved):
Looks good to me! I checked out the branch and it downloaded exactly as I'd expect without taking me to another page.

PR #5112, opened 2025-08-07
chore: migrates bot & agent views to standardized structure

techfg:
# What does this PR do?

Migrates `bot` and `agent` views to the standardized structure/layout.

Resolves the following issues:

1. Consistent UI/UX relative to other "detail" type views
2. When navigating the view, the record was in "READ" mode but the monaco editor was in "EDIT" mode
3. After saving changes, the Monaco editor remained in "EDIT" mode

# Testing

Manually tested and confirmed expected behavior.

PR #5113, opened 2025-08-07
fix: ensure setValue only called when legitimate change is occuring

techfg:
# What does this PR do?

With the changes made in #5111 & specifically #5112, it exposed an issue in the way we respond to `onChange` events from Monaco.  Those two PRs ensured that we are using Monaco with a controlled input and also ensured that we toggle the `mode` from READ->EDIT and vice-versa based on user actions.

The proper toggling of the mode resulted in Monaco firing the `onChange` event when changing back to READ mode even though the controlled input is responsible for the change.  Monaco is smart about not firing onChange when in EDIT mode but not when in READ mode.

This PR works around the Monaco issue by detecting the current mode and old/new value and only calling `setValue` when in EDIT mode and the value is actually different than what we have in state (meaning the user typed something in the editor).

[Monaco Detection in EDIT mode](https://github.com/suren-atoyan/monaco-react/blob/1e3c3efc29ae3b6d07af7545a9a3da6fa9142ba4/src/Editor/Editor.tsx#L213)

[Monaco in READ mode](https://github.com/suren-atoyan/monaco-react/blob/1e3c3efc29ae3b6d07af7545a9a3da6fa9142ba4/src/Editor/Editor.tsx#L99)

# Testing

Manually tested and confirmed.

### DIFF
--- a/libs/apps/uesio/io/bundle/componentpacks/main/src/utilities/codefield/codefield.tsx
+++ b/libs/apps/uesio/io/bundle/componentpacks/main/src/utilities/codefield/codefield.tsx
@@ -1,8 +1,8 @@
 import { component, definition, styles, api, context, util } from "@uesio/ui"
-import Editor, { loader, Monaco } from "@monaco-editor/react"
+import Editor, { loader, Monaco, type OnChange } from "@monaco-editor/react"
 import type monaco from "monaco-editor"
 import type { CodeFieldUtilityProps } from "./types"
-import { useEffect, useMemo, useState } from "react"
+import { useEffect, useMemo, useRef, useState } from "react"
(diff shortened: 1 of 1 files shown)


humandad:
✅  I wonder if this may fix the bug with not being able to update to the latest version of monaco?

techfg:
> ✅ I wonder if this may fix the bug with not being able to update to the latest version of monaco?

I suspect that it won't because in prior to the recent work in/around this area, the editor was always staying in "EDIT" mode and not toggling to read mode.  That said, we moved to a "true" controlled input in #5111 so that might help the monaco scenario.

Either way, it's worth a shot - I was never really able to repro the monaco thing locally, can you give it a shot with current code in `main`?

PR #5118, opened 2025-08-08
fix: enforce backend create domain feature flag

techfg:
# What does this PR do?

Enforces the `manage_domains` feature flag on the backend.

Previously, the "domains" view did not offer "create domain" functionality but the backend did not enforce it meaning direct API calls would succeed in creating domain even if user did not have feature flag enabled.

# Testing

Tested manually and confirmed.


humandad:
✅

PR #5119, opened 2025-08-09
fix: ensure collectionId of user file matches collection of recordid

techfg:
# What does this PR do?

Fixes an issue where on 2nd (and subsequent) uploads of a userfile to and attachment, the `collection` would be `userfile` instead of the `id` of the collection that "owns" the userfile.  Additionally, adds server side validation to ensure that a userfile can never have a collection of `userfile` and adds toast notifications to `uploadFile` and `deleteFile` to surface errors to user in UI.

In [this commit](https://github.com/ues-io/uesio/commit/ade91a8913cd23735d86da29cd3270d70a35efc7#diff-55d7c9ea51b4f7299ecf6ba9cb00a07c2970652389e1e79b334a5eb303cb2f5dL24), the `collectionid` value was changed to use the `$Collection{id}`, however when there is a record in context for `userfile`, the collection is now `userfile`.

@humandad  - The hierarchy and flow of these components is a little odd in my opinion.  In short, [section_attachments](https://github.com/ues-io/uesio/pull/5119/files#diff-88f1f4feb81702fb98eadab44492744c30f50d3efecf1e61bd85536e408cdaa0R34), which is a component, directly uses `$Param`. Should components ever use `param` and should they only use `prop` and any view should transpose its params to props?  Seems like only views should ever use `params` as I don't believe there is a way to define `params` on a component in its definition.  As it stands, the hierarchy of all these components does not have complete definition of props/params so that needs to be adjusted, just want to confirm how to adjust.  This is not directly related to this PR, but indirectly.

# Testing

Manually tested `Files` & "Collection Record Data Management" (the only two places where this functionality is leveraged) to ensure that files are added as expected.

### DIFF
--- a/apps/platform/pkg/bot/systemdialect/systembot_after_userfile.go
+++ b/apps/platform/pkg/bot/systemdialect/systembot_after_userfile.go
@@ -2,6 +2,7 @@ package systemdialect
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/thecloudmasters/uesio/pkg/auth"
@@ -15,6 +16,7 @@ const (
 	userCollectionId       = "uesio/core.user"
 	studioFileCollectionId = "uesio/studio.file"
 	studioBotCollectionId  = "uesio/studio.bot"
+	userFileCollectionId   = "uesio/core.userfile"
 )
 
 func runUserFileAfterSaveBot(ctx context.Context, request *wire.SaveOp, connection wire.Connection, session *sess.Session) error {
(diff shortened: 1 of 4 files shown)


techfg:
Just to close the loop on the question regarding $Param/$Prop:

1. Only views should have Params, component should not
2. Only components should have Props, views should not

The situation with [section_attachments](https://github.com/ues-io/uesio/pull/5119/files#diff-88f1f4feb81702fb98eadab44492744c30f50d3efecf1e61bd85536e408cdaa0R29) is one in which a component is actually calling a view so its a direct pass through (if they exist on the parent view(s)).  There is something to be said for having section_components receive props for these and then having it translate the props to params for the `file_attachments` view, however this is a consistent pattern used currently so if it's determined that it should change, which there is merit to doing so, it should be across the board and not a one-off change.  Given this, for now, leaving as-is but this should be re-evaluated broadly and a decision made on which approach is best and adjusted accordingly.

